PR #19, opened 2015-02-24
Add the Korean language.

dorajistyle:
Thanks for the great I18N Go package.
I've just added the Korean language.


nicksnyder:
Thanks!

PR #20, opened 2015-02-26
Added Turkish

udokmeci:
PluralSpecs and test
tested OK.

### DIFF
--- a/i18n/language/pluralspec.go
+++ b/i18n/language/pluralspec.go
@@ -339,6 +339,17 @@ var pluralSpecs = map[string]*PluralSpec{
 		},
 	},
 
+	// Turkish
+	"tr": &PluralSpec{
+		Plurals: newPluralSet(Zero, Other),
+		PluralFunc: func(ops *operands) Plural {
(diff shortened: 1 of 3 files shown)


i18n/language/pluralspec.go
```diff
@@ -339,6 +339,17 @@ var pluralSpecs = map[string]*PluralSpec{
 		},
 	},
 
+	// Turkish
+	"tr": &PluralSpec{
+		Plurals: newPluralSet(Zero, Other),
```
nicksnyder: This logic doesn't seem consistent with:
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html

Can you double check this?

udokmeci:
Ok now it fits

nicksnyder:
I still don't think this logic is correct. You haven't tested numbers like "1.0", "1.000" or "1.00001".

Please look at the implementation and tests for Spanish; they should be identical for Turkish.

udokmeci:
Sorry for too much try. I used right operands this time and added the ones you mention to tests.

PR #21, opened 2015-04-30
Add Malay `ms` and Indonesian `id`

sogko:
Malay plural rules:
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#ms

Indonesian plural rules:
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#id


i18n/language/pluralspec.go
```diff
@@ -242,6 +250,14 @@ var pluralSpecs = map[string]*PluralSpec{
 		},
 	},
 
+	// Malay
+	"ms": &PluralSpec{
+		Plurals: newPluralSet(Other),
+		PluralFunc: func(ops *operands) Plural {
+			return Other
```
nicksnyder: It looks like you are missing the One category for Malay

sogko: Thanks for taking your time to look at this.

I just double-checked, the cardinal rule for [Malay `ms`](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#ms) only specified `other`, i.e. the plural form is the same as its singular form.

nicksnyder: Oh yeah, I mixed up cardinal and ordinal.

PR #22, opened 2015-05-01
Export SupportedLanguage

mash:
I'd like to know which language go-i18n decided to use among the candidates caller provided to Tfunc, for example save the language into database for pushing localized content afterwards.

### DIFF
--- a/i18n/bundle/bundle.go
+++ b/i18n/bundle/bundle.go
@@ -137,15 +137,7 @@ func (b *Bundle) MustTfunc(languageSource string, languageSources ...string) Tra
 //
 // It can parse languages from Accept-Language headers (RFC 2616).
 func (b *Bundle) Tfunc(src string, srcs ...string) (TranslateFunc, error) {
-	lang := b.translatedLanguage(src)
-	if lang == nil {
-		for _, src := range srcs {
-			lang = b.translatedLanguage(src)
-			if lang != nil {
-				break
-			}
-		}
-	}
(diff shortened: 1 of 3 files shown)


mash:
OK, I did invert the condition.
Also, it was already gofmt-ed :)

nicksnyder:
Thanks for this contribution!

In your use case, do you expect to call Tfunc and SupportedLanguage with the same frequency (e.g. every request)?

mash:
Yes, so it's best if MustTfunc (called with a fallback language so that it doesn't panic) returned both the TranslateFunc and Language, but I thought changing the interface is not an option.

nicksnyder:
Ok. Let me think about if we should break API compatibility.

mash:
or maybe just add two more functions (MustTfuncAndLanguage and TfuncAndLanguage)? either is fine for me

nicksnyder:
I like the idea of MustTfuncAndLanguage and TfuncAndLanguage. Would you be able to update your pull request?

mash:
OK, I'll work on that

nicksnyder:
Sorry for the delay, I didn't realize you had updated the pull request. I merged your commit with some minor changes. I kept supportedLanguages unexported as it seemed redundant given the new API. Please let me know if you have a reason for it needing to still be exported and I and easily export it.

Thanks again for your contribution!

mash:
Thanks!
I'm fine with keeping supportedLanguages unexported.

PR #25, opened 2015-07-10
bundle: runtime-reflect structs to map[string]interface{}

### LINKED ISSUE #24: Support structs instead of map[string]interface{}
Hey @nicksnyder!

Thanks for writing this great series of packages. We're looking into using them and we're running into a fairly large issue: when we call a `Tfunc`, it is automatically cast to a `map[string]interface{}` instead of being left alone. As a result, my struct doesn't resolve properties.

Thoughts?

parkr:
Fixes #24.

@nicksnyder The `structs` package is quite lightweight. It is also licensed under MIT. I can attempt to extract the `Struct.structFields()` and `Struct.Map()` methods if you would prefer.

### DIFF
--- a/i18n/bundle/bundle.go
+++ b/i18n/bundle/bundle.go
@@ -10,6 +10,7 @@ import (
 
 	"path/filepath"
 
+	"github.com/fatih/structs"
 	"github.com/nicksnyder/go-i18n/i18n/language"
 	"github.com/nicksnyder/go-i18n/i18n/translation"
 )
@@ -237,7 +238,7 @@ func (b *Bundle) translate(lang *language.Language, translationID string, args .
 
 	var data map[string]interface{}
 	if len(args) > 0 {
-		data, _ = args[0].(map[string]interface{})
+		data = toMap(args[0])
 	}
 
 	if isNumber(count) {
@@ -261,4 +262,13 @@ func isNumber(n interface{}) bool {
 		return true
 	}
 	return false
+}
+
+func toMap(input interface{}) (data map[string]interface{}) {
+	if structs.IsStruct(input) {
+		data = structs.Map(input)
+	} else {
+		data, _ = input.(map[string]interface{})
+	}
+	return
 }
--- a/i18n/bundle/bundle.go
+++ b/i18n/bundle/bundle.go
@@ -5,12 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"reflect"
 
 	//	"launchpad.net/goyaml"
 
 	"path/filepath"
 
-	"github.com/fatih/structs"
 	"github.com/nicksnyder/go-i18n/i18n/language"
 	"github.com/nicksnyder/go-i18n/i18n/translation"
 )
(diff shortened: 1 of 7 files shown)


parkr:
@nicksnyder I believe this will do it.

i18n/bundle/bundle.go
```diff
@@ -262,3 +263,18 @@ func isNumber(n interface{}) bool {
...
+	switch v.Kind() {
+	case reflect.Map:
+		data, _ := input.(map[string]interface{})
+		return data
+	case reflect.Ptr:
+		return toMap(v.Elem().Interface())
+	case reflect.Struct:
+		return structToMap(v)
+	default:
+		panic(fmt.Errorf("i18n: cannot handle type %s", v.Kind()))
```
nicksnyder: As a library, I would prefer not to panic under any circumstances. We should just silently discard the data if it isn't a type we expect.

i18n/bundle/structs.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Panic if the value does not represent a struct.
+func assertStruct(v reflect.Value) {
+	if kind := v.Kind(); kind != reflect.Struct {
+		panic(fmt.Errorf("struct expected, got %s", kind))
```
nicksnyder: same thing here. let's do without the panic

nicksnyder:
Looks great, and thanks! Just a few comments about removing the panics.

Also, if you are feeling adventurous, it would be cool to have some benchmarks for structs and maps to see if there is a noticeable difference.

parkr:
> Also, if you are feeling adventurous, it would be cool to have some benchmarks for structs and maps to see if there is a noticeable difference.

Sure, I could do that.

parkr:
Benchmarks show this:

```text
~/code/go-i18n/i18n/bundle#struct-to-map$ go test -bench=. -cpu=1,2
PASS
BenchmarkToMapWithMap	100000000	        18.8 ns/op
BenchmarkToMapWithMap-2	100000000	        18.6 ns/op
BenchmarkToMapWithStruct	 1000000	      1693 ns/op
BenchmarkToMapWithStruct-2	 1000000	      1340 ns/op
BenchmarkStructToMapWithMap	 2000000	       692 ns/op
BenchmarkStructToMapWithMap-2	 2000000	       737 ns/op
BenchmarkStructToMapWithStruct	 1000000	      1533 ns/op
BenchmarkStructToMapWithStruct-2	 1000000	      1234 ns/op
ok  	_/Users/parker/code/go-i18n/i18n/bundle	13.952s
```

parkr:
@nicksnyder Let me know what you think. Would love to get this moving along sooner rather than later. :)

parkr:
Oh wow! Thank you!

nicksnyder:
Thanks again for your contribution @parkr !

I did a bit of code simplification in a268f4d6ed86ee49fff2665174c2d22885290491 as well as create a benchmark for Tfunc itself, as that is more interesting to measure than the raw performance of toMap.

```
BenchmarkTranslateWithMap		500000		2696 ns/op
BenchmarkTranslateWithStruct	300000		4010 ns/op
```

This shows us that using a struct has a 50% performance penalty in the best case (it gets worse the larger the struct is), but is still sufficiently fast in general.

parkr:
Great work, @nicksnyder. Thanks for tidying it up. Interesting benchmark – I wonder if we can parallelize the processing of the fields instead of blocking. Think that would help? Overhead of the channel might make it less efficient, but who knows.

An extra 2 microseconds isn't _too_ bad. Does the 50% performance penalty grow as the struct field number increases (i.e. 75% for a field with 10 fields, 90% hit for 100+ fields, etc)?

nicksnyder:
I did a few spot tests and it increased by about 100ns per field on the root struct.

I just realized that we can (and should) skip this conversion entirely for the non-plural case, since there is no need to add any data. I can work on this change.

In any given page/request (assuming a web context) you probably won't have more than ~10 pluralized strings so the perf penalty seems insignificant in the grand scheme of things (web latency is measured in milliseconds, not microseconds).

parkr:
> I just realized that we can (and should) skip this conversion entirely for the non-plural case, since there is no need to add any data. I can work on this change.

Did you end up making this change, @nicksnyder? Not too too worried, but I nevertheless feel the need for speed. :racehorse:

nicksnyder:
Thanks for reminder. I had the changes locally but forgot to push.

parkr:
Brilliant. Thanks again!!

PR #26, opened 2015-07-26
Add Norwegian Bokmål

AmandaCameron:
Sorry for the huge number of changes. go-fmt had something to say about the rest of the code. :s

### DIFF
--- a/i18n/language/pluralspec.go
+++ b/i18n/language/pluralspec.go
@@ -161,7 +161,7 @@ var pluralSpecs = map[string]*PluralSpec{
 	"no-nb": &PluralSpec{
 		Plurals: newPluralSet(One, Other),
 		PluralFunc: func(ops *operands) Plural {
-			if ops.I == 1 && ops.V == 0 {
+			if ops.I == 1 && ops.W == 0 {
 				return One
 			}
 			return Other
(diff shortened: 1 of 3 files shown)


nicksnyder:
Thanks for running gofmt!

It looks like your implementation and tests are incorrect. Both should be identical to Spanish according to the [plural rules](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#nb).

AmandaCameron:
Whoops. I'll fix that, the problem I had was that the variables were rather un-untiutively named for someone comfonr from a zero-state of the CLDR.

nicksnyder:
Hey @AmandaCameron 

Sorry, this kind of dropped off my radar. It looks like Travis builds fail with your changes, do you mind rebasing and cleaning this up?

I also noticed that you added two more languages to the implementation but did not update the tests or README. Could you do this?

AmandaCameron:
Whoops, sorry, I'll have to get to it later, as I'm pretty busy with some
RL stuff right now, but I had sketched out a couple of other labnguages
after no-nb to toy with it some more, and forgot by the time I fgot to
fixing it up again for the PR.

Feel free to close this and I'll open a new one when I have more time to
work on this stuff.

i18n/language/pluralspec.go
```diff
@@ -266,6 +301,22 @@ var pluralSpecs = map[string]*PluralSpec{
 		},
 	},
 
+	// Macedonian
+	"mk": &PluralSpec{
```
parkr: macedonian?! neat.

i18n/language/pluralspec.go
```diff
@@ -258,6 +269,30 @@ var pluralSpecs = map[string]*PluralSpec{
 		},
 	},
 
+	// Latvian
+	"lv": &PluralSpec{
```
parkr: latvian! great idea.

nicksnyder:
Closing this since @parkr submitted these languages in #36 and #35

PR #29, opened 2015-08-17
pluralspec: add 'th' Thai support.

parkr:
Looks like it always uses `other`? 
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html

/cc @sayrer

### DIFF
--- a/i18n/language/pluralspec_test.go
+++ b/i18n/language/pluralspec_test.go
@@ -504,6 +504,25 @@ func TestUkrainian(t *testing.T) {
 	runTests(t, "uk", tests)
 }
 
+func TestThai(t *testing.T) {
+	tests := []pluralTest{
+		{0, Other},
+		{1, Other},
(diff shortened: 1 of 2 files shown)


i18n/language/pluralspec_test.go
```diff
@@ -504,6 +504,25 @@ func TestUkrainian(t *testing.T) {
 	runTests(t, "uk", tests)
 }
 
+func TestThai(t *testing.T) {
+	tests := []pluralTest{
```
nicksnyder: There are already languages that always return Other, so these tests should be identical to those (e.g. Chinese).

Perhaps it would be good to make a helper test function that tests languages which always return Other.

nicksnyder:
Actually, someone beat you to it today: #28

parkr:
Oh, wow. Thanks.

PR #33, opened 2015-09-24
i18n/language: add vietnamese pluralspec

parkr:
Hey, @nicksnyder! Adding Vietnamese (`vi-VN`, `vi`) in here. Used the pluralization rules from here: http://localization-guide.readthedocs.org/en/latest/l10n/pluralforms.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ Supported languages
 * Tigrinya (`ti`)
 * Turkish (`tr`)
 * Ukrainian (`uk`)
+* Vietnamese (`vi`)
 
 Adding new languages
 --------------------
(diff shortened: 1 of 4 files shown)


parkr:
Looks like a little gofmt snuck in there. Hope that's ok!

nicksnyder:
Thanks!

gofmt is always good; unformatted code always finds a way to sneak in as sometimes it is not obvious in the diff.

Is there any particular reason that you decided to use readthedocs instead of [CLDR](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)? I have been using the latter so I would like to maintain consistency going forward. I would hope they are identical, but you never know.

parkr:
@nicksnyder it's just what I found! Happy to use that Unicode.org link instead.

PR #35, opened 2015-09-24
i18n/language: add Macedonian

parkr:
Rules per http://localization-guide.readthedocs.org/en/latest/l10n/pluralforms.html

- [x] CLDR rules
- [x] Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ Supported languages
 * Japanese (`ja`)
 * Lithuanian (`lt`)
 * Malay (`ms`)
+* Macedonian (`mk`)
 * Norweigan (`no`)
 * Polish (`pl`)
 * Portuguese (`pt`)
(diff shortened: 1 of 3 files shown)


nicksnyder:
Please use [CLDR](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html) rules as source of truth. I am happy to merge after there are tests and merge conflicts are resolved.

Thanks!

parkr:
Hey, Nick! Hear you loud and clear. The [CLDR rules for `mk`](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#mk) seem to indicate a difference in calculating the proper `Plural` for cardinal, ordinal, and range, respectively. Which should I use here?

parkr:
Updated with Cardinal CLDR rules. Let me know if you would like me to make any further adjustments. Thanks!

nicksnyder:
Yes, cardinal is the correct ruleset to use.

PR #36, opened 2015-09-24
i18n/language: add Latvian

parkr:
- [x] [Latvian CLDR rules](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#lv)
- [x] Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Supported languages
 * Indonesian (`id`)
 * Italian (`it`)
 * Korean (`ko`)
+* Latvian (`lv`)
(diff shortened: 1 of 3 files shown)


nicksnyder:
Hmm...this seems to be inconsistent with [CLDR](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#lv).

I would like to stick with CLDR as I have aspirational goals to do code generation for these locales and tests based on CLDR raw data.

parkr:
Updated! Sorry about the confusion, Nick.

PR #38, opened 2015-09-29
Expose the loaded bundles for tests and verification purposes.

dennisfaust:


nicksnyder:
@dennisfaust 
Can you describe more what exactly you are trying to test? The data structure you are exposing is an implementation detail of the library that really shouldn't be part of the exported api.

dennisfaust:
It's more a matter of verification. We need to have tests that ensure all our translations are complete with respect to our canonical one which is en-us. I wrote tests to verify the number of translation keys in each, and then traversed the keys in en-us, verifying that all the other language files have a translation for each key. Interesting enough, even though our files aren't large running these tests found a translation that was in the en-us file that wasn't in the others.

If you'd prefer not to expose the bundles maybe a utility function which return the number of keys in each? And another which can compare the keys between translations?

nicksnyder:
@dennisfaust 

> Interesting enough, even though our files aren't large running these tests found a translation that was in the en-us file that wasn't in the others.

Out of curiosity, do you use the goi18n command? It can create a file that contains all translations which haven't been translated yet. If you find a bug, please let me know.

> If you'd prefer not to expose the bundles maybe a utility function which return the number of keys in each? And another which can compare the keys between translations?

I would prefer to have two simpler functions that you could use to fetch information about what is in the bundle, and then you can use the normal Tfunc interface to see if everything has a translation. Would this work for you?

```go
// LanguageTags returns the tags of all languages that that have been added.
func (b *Bundle) LanguageTags() []string {
	// implementation
}

// LanguageTranslationIDs returns the ids of all translations that have been added for a given language.
func (b *Bundle) LanguageTranslationIDs(languageTag string) []string {
	// implementation
}
```

Similarly in i18n.go
```go
// LanguageTags returns the tags of all languages that have been added.
func LanguageTags() []string {
	return defaultBundle.LanguageTags()
}

// LanguageTranslationIDs returns the ids of all translations that have been added for a given language.
func LanguageTranslationIDs(languageTag string) []string {
	return defaultBundle.LanguageTranslationIDs(languageTag)
}
```

dennisfaust:
Dear Nick,

    Those would suit just fine. I realize goi18n can create the file with
missing translations and that's useful but in our situation we like to have
our test suite be a failsafe that we don't forget or overlook something.
The ability to do it easily as a test is important to us. I'd be happy to
add these if you'd like and update the docs.

Thanks,

Dennis

nicksnyder:
@dennisfaust Yes please, if you could update your pull request with those changes, that would be helpful! Thanks!

i18n/i18n.go
```diff
@@ -56,7 +56,7 @@
 package i18n
 
 import (
-	"github.com/nicksnyder/go-i18n/i18n/bundle"
+	"github.com/dennisfaust/go-i18n/i18n/bundle"
 	"github.com/nicksnyder/go-i18n/i18n/language"
```
dennisfaust: This would need to be changed back when the merge happens.

i18n/i18n.go
```diff
@@ -56,7 +56,7 @@
 package i18n
 
 import (
-	"github.com/nicksnyder/go-i18n/i18n/bundle"
+	"github.com/dennisfaust/go-i18n/i18n/bundle"
```
parkr: TravisCI runs this within `$GOPATH/src/github.com/nicksnyder/go-i18n`, so this is actually unnecessary for CI purposes here.

i18n/bundle/bundle.go
```diff
@@ -129,9 +129,30 @@ func (b *Bundle) AddTranslation(lang *language.Language, translations ...transla
...
+}
+
+// LanguageTranslationIDs returns the ids of all translations that have been added for a given language.
+func (b *Bundle) LanguageTranslationIDs(languageTag string) []string {
+	langTrans := b.translations[languageTag]
+	if langTrans != nil {
+		iDs := make([]string, len(langTrans))
+		i := 0
+		for _, translation := range langTrans {
+			iDs[i] = translation.ID()
```
parkr: consider using `append()` here:

```go
iDs = append(iDs, translation.ID())
```

i18n/bundle/bundle.go
```diff
@@ -129,9 +129,30 @@ func (b *Bundle) AddTranslation(lang *language.Language, translations ...transla
...
 
-// Translations returns all translations in the bundle.
-func (b *Bundle) Translations() map[string]map[string]translation.Translation {
-	return b.translations
+// LanguageTags returns the tags of all languages that that have been added.
+func (b *Bundle) LanguageTags() []string {
+	tags := make([]string, len(b.translations))
+	i := 0
+	for k := range b.translations {
+		tags[i] = k
```
parkr: same here – consider using `append()` and maybe using a named return value to simplify things

```go
func (b *Bundle) LanguageTags() (tags []string) {
    for _, translation := range b.translations {
        tags = append(tags, translation)
    }
    return
}
```

i18n/bundle/bundle.go
dennisfaust: It would but I read somewhere that append was much slower if you know the size of the target and can just add it yourself.

i18n/bundle/bundle.go
dennisfaust: I think you're right. It's more readable that way.

dennisfaust:
@nicksnyder please have a look at this. I think it's done but I'd like to see what you think. 

Thanks!

i18n/bundle/bundle_test.go
```diff
@@ -2,10 +2,11 @@ package bundle
 
 import (
 	"fmt"
-	"testing"
-
```
nicksnyder: Please keep the empty line between standard library imports and other imports

i18n/i18n.go
```diff
@@ -86,6 +86,22 @@ func IdentityTfunc() TranslateFunc {
 
 var defaultBundle = bundle.New()
 
+// AllBundles - returns all the loaded translations. This is useful
+//  for tests and verifications.
+//func AllBundles() map[string]map[string]translation.Translation {
+//	return defaultBundle.Translations()
+//}
```
nicksnyder: Delete commented out code

i18n/bundle/bundle.go
```diff
@@ -134,6 +134,25 @@ func (b *Bundle) Translations() map[string]map[string]translation.Translation {
 	return b.translations
 }
 
+// LanguageTags returns the tags of all languages that that have been added.
+func (b *Bundle) LanguageTags() (tags []string) {
+	for k := range b.translations {
+		tags = append(tags, k)
+	}
+	return
```
nicksnyder: Please use explicit return statements

```go
return tags
````

File: i18n/bundle/bundle.go [hunk 8]
```diff
@@ -134,6 +134,25 @@ func (b *Bundle) Translations() map[string]map[string]translation.Translation {
...
+
+// LanguageTranslationIDs returns the ids of all translations that have been added for a given language.
+func (b *Bundle) LanguageTranslationIDs(languageTag string) (iDs []string) {
+	langTrans := b.translations[languageTag]
+	if langTrans != nil {
+		for _, translation := range langTrans {
+			iDs = append(iDs, translation.ID())
+		}
+	}
+	return
```
nicksnyder: Please use explicit return statements

```go
return ids
````

File: i18n/bundle/bundle.go [hunk 9]
```diff
@@ -134,6 +134,25 @@ func (b *Bundle) Translations() map[string]map[string]translation.Translation {
 	return b.translations
 }
 
+// LanguageTags returns the tags of all languages that that have been added.
+func (b *Bundle) LanguageTags() (tags []string) {
```
nicksnyder: Please don't use named return variables for functions that only have one return variable. I would rather see the variable declaration in the function implementation.

```go
var tags []string
```

i18n/bundle/bundle.go
```diff
@@ -134,6 +134,25 @@ func (b *Bundle) Translations() map[string]map[string]translation.Translation {
...
+	for k := range b.translations {
+		tags = append(tags, k)
+	}
+	return
+}
+
+// LanguageTranslationIDs returns the ids of all translations that have been added for a given language.
+func (b *Bundle) LanguageTranslationIDs(languageTag string) (iDs []string) {
```
nicksnyder: Please don't use named return variables for functions that only have one return variable. I would rather see the variable declaration in the function implementation.

Also, "ids" not "iDs"

```go
var ids []string
```

i18n/bundle/bundle.go
```diff
@@ -134,6 +134,25 @@ func (b *Bundle) Translations() map[string]map[string]translation.Translation {
...
+	langTrans := b.translations[languageTag]
+	if langTrans != nil {
```
nicksnyder: It is safe to range over a nil map, so this guard is unnecessary

i18n/bundle/bundle.go
```diff
@@ -134,6 +134,25 @@ func (b *Bundle) Translations() map[string]map[string]translation.Translation {
...
+		for _, translation := range langTrans {
```
nicksnyder: for simplicity, we can just iterate over the keys and collect those, right?

```go
for id := range langTrans {
  ids = append(ids, id)
}
```

i18n/bundle/bundle_test.go
```diff
@@ -33,6 +34,28 @@ func TestMustTfunc(t *testing.T) {
...
+	spanishLanguage := languageWithTag("es")
+	addFakeTranslation(t, b, englishLanguage, "English"+translationID)
+	addFakeTranslation(t, b, frenchLanguage, translationID)
+	addFakeTranslation(t, b, spanishLanguage, translationID)
+
+	tags := b.LanguageTags()
+	sort.Strings(tags)
+	compareTo := []string{englishLanguage.Tag, spanishLanguage.Tag, frenchLanguage.Tag}
+	if !reflect.DeepEqual(tags, compareTo) {
+		t.Error("Problem encountered with LanguageTags.")
```
nicksnyder: Please create a more descriptive error message

```go
t.Errorf("LanguageTags() = %#v; expected %#v", tags, expectedTags)
```

i18n/bundle/bundle_test.go
```diff
@@ -33,6 +34,28 @@ func TestMustTfunc(t *testing.T) {
...
+	}
+
+	if b.LanguageTranslationIDs(englishLanguage.Tag)[0] != "English"+translationID {
+		t.Error("Problem with LanguageTranslationIDs")
```
nicksnyder: Please create a more descriptive error message

```go
t.Errorf("LanguageTranslationIds() = %#v; expected %#v", ids, expectedIds)
```

i18n/bundle/bundle_test.go
```diff
@@ -33,6 +34,28 @@ func TestMustTfunc(t *testing.T) {
...
+	addFakeTranslation(t, b, spanishLanguage, translationID)
+
+	tags := b.LanguageTags()
+	sort.Strings(tags)
+	compareTo := []string{englishLanguage.Tag, spanishLanguage.Tag, frenchLanguage.Tag}
+	if !reflect.DeepEqual(tags, compareTo) {
+		t.Error("Problem encountered with LanguageTags.")
+	}
+
+	if b.LanguageTranslationIDs(englishLanguage.Tag)[0] != "English"+translationID {
```
nicksnyder: I think you should do the same thing you did above and use an array instead of just checking the first element in the array. That way we also test there aren't additional elements we aren't expecting.

nicksnyder:
@dennisfaust Thanks! I added some comments to the diff

i18n/bundle/bundle.go
parkr: I asked him to do that -- I figured it was much cleaner, and that it was essentially doing that for you. Cleaner code in my opinion but I defer:

nicksnyder: Sorry, I didn't realize you had already made some comments (I kind of hate how comments disappear after a new diff is uploaded).

In any case, I think it is best to avoid named return values except for a few circumstances.
  1. Your function returns multiple values of the same type so you need a name to disambiguate.
  2. Your function uses defer to modify return values before actually returning.

Even when I have named return values due to above exceptions, I would always still use explicit return statements because it is easier to read.

In this particular case, the named return value just unnecessarily stutters the method signature.

dennisfaust:
I believe I have integrated all your comments and suggestions. Thanks for the link to the code review comments. I'm going to bring those suggestions into our projects as well.

nicksnyder:
@dennisfaust Code looks good. I have one last request before I merge. Can you please squash all of these commits into one commit with a nice commit message?

```
git reset --soft HEAD~12
git commit
```

dennisfaust:
@nicksnyder Done.

nicksnyder:
@dennisfaust I still see 12 commits here. You will need to force push to your fork

PR #41, opened 2015-11-17
added yaml support with tests

zippoxer:
I believe my translators would be more comfortable with YAML due to it's noiseless syntax, so I added support for it.

I also changed one of the test files used by the merge test from JSON to YAML, to make sure we can merge YAML as well.

I also added an example file identical to the original with the sole exception that it parses a YAML file.

### DIFF
--- a/goi18n/merge_test.go
+++ b/goi18n/merge_test.go
@@ -7,8 +7,7 @@ import (
 	"testing"
 )
 
-func TestMergeExecute(t *testing.T) {
-	resetDir(t, "testdata/output")
+func TestMergeExecuteJSON(t *testing.T) {
 	files := []string{
 		"testdata/input/en-us.one.json",
 		"testdata/input/en-us.two.json",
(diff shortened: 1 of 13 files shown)


zippoxer:
BTW, the import paths are changed to `github.com/zippoxer/...`. I decided to keep it this way in case the author wants to test this before merging.

goi18n/merge_test.go
```diff
@@ -10,7 +10,7 @@ import (
 func TestMergeExecute(t *testing.T) {
 	resetDir(t, "testdata/output")
 	files := []string{
-		"testdata/input/en-us.one.json",
+		"testdata/input/en-us.one.yaml",
```
nicksnyder: Instead of editing this test, can you leave this test as-is and add a new test which uses the new yaml file? Basically copy/paste.

i18n/translation/translation.go
```diff
@@ -44,27 +45,39 @@ func NewTranslation(data map[string]interface{}) (Translation, error) {
 			return nil, err
 		}
 		return &singleTranslation{id, tmpl}, nil
-	case map[string]interface{}:
-		templates := make(map[language.Plural]*template, len(translation))
+	case map[interface{}]interface{}:
+		// The YAML parser uses interface{} keys.
```
nicksnyder: Can you make this comment more detailed?
"The YAML parser uses interface{} keys so first we convert them to string keys."

zippoxer:
@nicksnyder 

I applied the two changes in your comments, and had to force push so the squashed commit reads like it has been committed 4 days ago.

parkr:
This looks good to me.

PR #46, opened 2016-01-11
Implement translation namespaces

gbitten:
This is more a hack than a PR. Anyway, the patch works really well for me, so I am showing it here
It creates a new package, nsi18n. This is a variant of i18n package and uses the bundle package without any modification. The translations can be stored in different namespaces. An application can define which namespaces will be used when calls Tfunc, MustTfunc, MustTfuncAndLanguage or TfuncAndLanguage functions.

### DIFF
--- a/nsi18n/nsi18n.go
+++ b/nsi18n/nsi18n.go
@@ -95,7 +95,9 @@ var defaultbundles = make(map[string]*bundle.Bundle)
 // CreateNamespace creates translation namespaces.
 func CreateNamespace (namespaces ...string) {
 	for _, namespace := range namespaces {
-		defaultbundles[namespace] = bundle.New()
+		if defaultbundles[namespace] == nil {
+			defaultbundles[namespace] = bundle.New()
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


nicksnyder:
Can you explain the use-case for namespacing? It seems like you could just prefix translation keys with a string to accomplish the same thing.

gbitten:
I am building a modular web application heavily based on text where there
are many translations in each module. So, I create a global namespace with
translations used in the whole app and local namespaces for each module. As
you said, I could prefix the translation keys, but the keys in my app have
became longer and semantically complex than I would like. That is why I am
trying to avoid the prefix solution.

nicksnyder:
@gbitten (sorry for the delayed response)

I am glad you are able to use the library to solve the problem you have. That said, I am not sure this code belongs in the core library as it is. There appears to be a lot of duplicated API and implementation.

I am wonder if namespacing is the best solution to this problem though. Maybe there is an opportunity here to support fallback bundles generically. For example, each module in your app could have its own bundle and then fallback to the "common" bundle if a particular translation is not found.

PR #49, opened 2016-03-18
handle Count field from template

benoittoulme:


i18n/example_test.go
```diff
@@ -22,6 +23,7 @@ func Example() {
 	fmt.Println(T("your_unread_email_count", 1))
 	fmt.Println(T("your_unread_email_count", 2))
 	fmt.Println(T("my_height_in_meters", "1.7"))
+	fmt.Println(T("my_height_in_meters", countStruct))
```
nicksnyder: Instead of this test, can you please test 0,1,2 to make sure that pluralization is working correctly? Also it would be good to test both structs and maps.

```go
type Count struct { Count int }
fmt.Println(T("your_unread_email_count", Count{0}))
fmt.Println(T("your_unread_email_count", Count{1}))
fmt.Println(T("your_unread_email_count", Count{2}))

fmt.Println(T("your_unread_email_count", map[string]string{"Count": 0}))
fmt.Println(T("your_unread_email_count", map[string]string{"Count": 1}))
fmt.Println(T("your_unread_email_count", map[string]string{"Count": 2}))
```

nicksnyder:
Thanks for the suggestion @benoittoulme

I just added one comment on some additional tests to add.

benoittoulme:
I will do this when I have time.
BTW, is there any reason why the tests are not written using https://golang.org/pkg/testing/ ?
If you want I can take care of this as well, it is a simple change.

nicksnyder:
@benoittoulme There are tests that use the testing package.
```
$ find . -name *_test.go
./goi18n/merge_test.go
./i18n/bundle/bundle_test.go
./i18n/example_test.go
./i18n/exampletemplate_test.go
./i18n/language/language_test.go
./i18n/language/operands_test.go
./i18n/language/plural_test.go
./i18n/language/pluralspec_gen_test.go
./i18n/language/pluralspec_test.go
./i18n/translation/plural_translation_test.go
./i18n/translation/template_test.go
./i18n/translation/translation_test.go
```
`example_test.go` is a test as well as an example that appears in [godoc](https://godoc.org/github.com/nicksnyder/go-i18n/i18n).

I want certain things to be in the example files so developers know how they can call the function. If it is tested in the example, it doesn't necessarily need to be tested elsewhere, but you are welcome to add any tests that you think would be helpful to `bundle_test.go` (or any other file).

Which reminds me, could you also update the documentation for `TranslateFunc` in `i18n.go` to reference the new behavior? The wording looks like it is going to get more complicated.

mh-cbon:
hi, Is it possible to finish this PR ? seems useful.

Suggest something like this,
```
// TranslateFunc returns the translation of the string identified by translationID.
//
// If there is no translation for translationID, then the translationID itself is returned.
// This makes it easy to identify missing translations in your app.
//
// If translationID is a non-plural form, then the first variadic argument may be a map[string]interface{}
// or struct that contains template data.
//
// If translationID is a plural form, then the first variadic argument 
// must be a map[string]interface{} or a struct with a Count field, 
// or an integer type (int, int8, int16, int32, int64) or a float formatted as a string (e.g. "123.45").
// The second variadic argument may be a map[string]interface{} or struct that contains template data.
//
///*
//[
//  {
//    "id": "d_days",
//    "translation": {
//      "one": "{{.Count}} day",
//      "other": "{{.Count}} days"
//    }
//  },
//}
//*/
// Get plural form with a map[string]interface{} or a struct with a Count field.
//     T("d_days", map[string]interface{}{
//         "Count":    3,
//     })
///* Prints: 3 days */
//
// Get plural form with a first variadic argument as an integer type
//     T("d_days", 3)
///* Prints: 3 days */
```
unsure if examples are allowed here.

nicksnyder:
@benoittoulme Any plans to finish this or can someone else take over? iirc we just need updates to doc and some test cases.

@mh-cbon I would like any example code to be built and tested using this technique:
https://blog.golang.org/examples. For example: [example-test.go](https://github.com/nicksnyder/go-i18n/blob/master/i18n/example_test.go).

mh-cbon:
@nicksnyder i propose I open a new pr based on @benoittoulme works and your feedbacks. Ping me otherwise.

nicksnyder:
@mh-cbon Yes, feel free to open a new PR

nicksnyder:
This implementation was finished in #53

Issue #50, opened 2016-09-20
[Feature] Command goi18n generate dynamic struct R.go 


rodrigocorsi2:
Working with Android/Java the IDE generates a dynamic class `R.java` with all strings ids, this method avoid a lot of bugs. In a big project is easy to reference an invalid id or to have a lot of unused ids.

What do you think the command `goi18n` generates `R.go` ?
Example:

en-US.all.json
```json
[
  {
    "id": "settings_title",
    "translation": "Settings"
  }
]
```

```bash
$ goi18n -genconst en-US.all.json -outdir ./
```
this command generates `R.go`
```go
package R
const SettingsTitle string = "settings_title"
```

In code the new way will be:
```go
T(R.SettingsTitle)
```
This is only a suggestion, let me know what do you think.
I can help with this change
thanks

nicksnyder:
Sounds like a good idea to me! I will happily merge a PR that implements this.

Can you clarify if you think this should be a separate command or just something that happens (optionally) when goi18n is run for merging purposes?

Other thoughts

(1) Should it be possible to configure the name of "R" to be something else? R seems like a good default.
(2) Should this be opt-out instead of opt-in?

rodrigocorsi2:
Due to compatibility with old projects, I think it should be optional using some parameters on the command goi18n.
```
$ goi18n -genconst -package=R en-US.all.json -outdir ./
```
The user can put this command in a Makefile for example.

But if you prefer more solid approach we can change `translationID` to `int`
```go
type TranslateFunc func(translationID int, args ...interface{}) string
```
In this situation the command will generate "R" like this:
```go
package R
const SettingsTitle int = 0
const OtherString  int = 1
```

nicksnyder:
It is still unclear to me if you are proposing a separate command or not.

Would `$ goi18n -genconst -package=R en-US.all.json -outdir ./` also do the normal merge process or would it *only* generate the constants?

I don't want to change the type of translationID (it should remain a string).

rodrigocorsi2:
I'm not proposing to create a new command but add new parameters to `goi18n`.
The command `$ goi18n -genconst -package=R en-US.all.json -outdir ./` will only generate the constants.

nicksnyder:
Ok, the way I read it you are proposing a new command (in the same way that `git commit` is a different command than `git push`). goi18n === git, except right now it only has one implicit command (imagine if just running `git` did something useful).

There are two ways we could proceed:

[...]

PR #51, opened 2016-09-21
Command goi18n generate dynamic struct R.go #50

rodrigocorsi2:
Please let me know what do you think

I tried to put the translation in comments this way:
```go
// PersonGreeting id:person_greeting
// Hello {{.Person}}
const PersonGreeting string = "person_greeting"
```
But I haven't found a method of making. If you can, please help me to put the translation in the comments.

### DIFF
--- a/goi18n/constants_command.go
+++ b/goi18n/constants_command.go
@@ -1,15 +1,21 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"text/template"
+	"unicode"
 
 	"github.com/nicksnyder/go-i18n/i18n/bundle"
+	"github.com/nicksnyder/go-i18n/i18n/language"
+	"github.com/nicksnyder/go-i18n/i18n/translation"
 )
 
 type constantsCommand struct {
@@ -19,8 +25,9 @@ type constantsCommand struct {
 }
 
 type templateConstants struct {
-	ID   string
-	Name string
+	ID      string
+	Name    string
+	Comment string
 }
 
 type templateHeader struct {
@@ -34,6 +41,7 @@ var constTemplate = template.Must(template.New("").Parse(`// DON'T CHANGE THIS F
 
 package {{.PackageName}}
 {{range .Constants}}
+// {{.Name}} {{.Comment}}
 const {{.Name}} = "{{.ID}}"
 {{end}}`))
 
@@ -69,9 +77,10 @@ func (cc *constantsCommand) execute() error {
 	for i, id := range keys {
 		tmpl.Constants[i].ID = id
 		tmpl.Constants[i].Name = toCamelCase(id)
+		tmpl.Constants[i].Comment = toComment(lang[id])
 	}
 
-	filename := filepath.Join(cc.outdir, fmt.Sprintf("%s.go", cc.packageName))
+	filename := filepath.Join(cc.outdir, cc.packageName+".go")
 	f, err := os.Create(filename)
 	if err != nil {
 		return fmt.Errorf("failed to create file %s because %s", filename, err)
@@ -84,6 +93,56 @@ func (cc *constantsCommand) execute() error {
 	}
 
 	return nil
+}
+
+func (cc *constantsCommand) parse(arguments []string) {
+	constantsCmd := flag.NewFlagSet("constants", flag.ExitOnError)
+	constantsCmd.Usage = usageConstants
+
+	packageName := constantsCmd.String("package", "R", "")
+	outdir := constantsCmd.String("outdir", ".", "")
+
+	constantsCmd.Parse(arguments)
+
+	if constantsCmd.Parsed() {
+		cc.translationFiles = constantsCmd.Args()
+		cc.packageName = *packageName
+		cc.outdir = *outdir
+	} else {
+		usageConstants()
+	}
+}
+
+func (cc *constantsCommand) SetArgs(args []string) {
+	cc.translationFiles = args
+}
+
+func usageConstants() {
+	fmt.Printf(`Generate constant file from translation file.
+
+Usage:
+
+    goi18n constants [options] [file]
+
+Translation files:
+
+    A translation file contains the strings and translations for a single language.
+
+    Translation file names must have a suffix of a supported format (e.g. .json) and
+    contain a valid language tag as defined by RFC 5646 (e.g. en-us, fr, zh-hant, etc.).
+
+Options:
+
+    -package name
+        goi18n generates the constant file under the package name.
+        Default: R
+
+    -outdir directory
+        goi18n writes the constant file to this directory.
+        Default: .
+
+`)
+	os.Exit(1)
 }
 
 // commonInitialisms is a set of common initialisms.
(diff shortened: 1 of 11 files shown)


goi18n/goi18n.go
```diff
@@ -6,12 +6,89 @@ import (
...
+		if err := cc.execute(); err != nil {
+			fmt.Println(err.Error())
+			os.Exit(1)
+		}
+	}
+}
+
 func usage() {
-	fmt.Printf(`goi18n formats and merges translation files.
+	fmt.Printf(`goi18n tools for translation files.
```
nicksnyder: goi18n manages translation files

goi18n/goi18n.go
```diff
@@ -6,12 +6,89 @@ import (
...
 
 Usage:
 
-    goi18n [options] [files...]
+    goi18n merge     [options] [files...]
```
nicksnyder: It would be nice to have a brief (one sentence description) here for each command.

goi18n/testdata/expected/R.go
```diff
@@ -0,0 +1,26 @@
+// DON'T CHANGE THIS FILE MANUALLY
+// This file was generate using the command:
+// $ goi18n constants
+
+package R
+
+// DDays id:d_days
```
nicksnyder: This comment doesn't seem very useful to me since it just repeats what is in the line of code below. I would remove it.

goi18n/testdata/expected/R.go
```diff
@@ -0,0 +1,26 @@
...
+const DDays string = "d_days"
+
+// MyHeightInMeters id:my_height_in_meters
+const MyHeightInMeters string = "my_height_in_meters"
```
nicksnyder: Is there a reason you are explicitly specifying the string type? I would just omit it.

```
const MyHeightInMeters = "my_height_in_meters"
```

goi18n/goi18n.go
```diff
@@ -6,12 +6,89 @@ import (
...
+	case "merge":
+		mergeCmd.Parse(os.Args[2:])
+	case "constants":
+		constantsCmd.Parse(os.Args[2:])
+	default:
+		mergeCmd.Parse(os.Args[1:])
+	}
+
+	if mergeCmd.Parsed() {
+		if len(constantsCmd.Args()) != 1 {
```
nicksnyder: this should be checked inside of execute

goi18n/goi18n.go
```diff
@@ -6,12 +6,89 @@ import (
...
+			sourceLanguageTag: *sourceLanguage,
+			outdir:            *outdir,
+			format:            *format,
+		}
+		if err := mc.execute(); err != nil {
+			fmt.Println(err.Error())
+			os.Exit(1)
+		}
+	} else if constantsCmd.Parsed() {
+		if len(constantsCmd.Args()) != 1 {
```
nicksnyder: this should be checked inside of execute

goi18n/goi18n.go
```diff
@@ -6,12 +6,89 @@ import (
...
+			usageMerge()
+		}
+
+		mc := &mergeCommand{
+			translationFiles:  mergeCmd.Args(),
+			sourceLanguageTag: *sourceLanguage,
+			outdir:            *outdir,
+			format:            *format,
+		}
+		if err := mc.execute(); err != nil {
```
nicksnyder: duplicate logic here and for merge command. I think we should create a command interface that both commands implement.

rodrigocorsi2:
@nicksnyder all changes are done.
I think we need to put the comment on constant for two things:
1 - `golint` shows a warning with public constants without comment
2 - Some editors helps the programmer with a tooltip with content of the comment

goi18n/goi18n.go
```diff
@@ -6,12 +6,82 @@ import (
...
+
+For more details execute:
+
+    goi18n [command] -help
+
+`)
+	os.Exit(1)
+}
+
+func usageMerge() {
```
nicksnyder: Can you move this to be defined in the file where the merge command is defined? Same with `usageConstants()`

rodrigocorsi2:
@nicksnyder done

goi18n/constants_command.go
```diff
@@ -86,6 +95,56 @@ func (cc *constantsCommand) execute() error {
...
+func (cc *constantsCommand) parse(arguments []string) {
+	constantsCmd := flag.NewFlagSet("constants", flag.ExitOnError)
+	constantsCmd.Usage = usageConstants
+
+	packageName := constantsCmd.String("package", "R", "")
+	outdir := constantsCmd.String("outdir", ".", "")
+
+	constantsCmd.Parse(arguments)
+
+	if constantsCmd.Parsed() {
```
nicksnyder: this will always be true, right?

goi18n/merge_command.go
```diff
@@ -78,6 +81,30 @@ func (mc *mergeCommand) execute() error {
...
+	mergeCmd := flag.NewFlagSet("merge", flag.ExitOnError)
+	mergeCmd.Usage = usageMerge
+
+	sourceLanguage := mergeCmd.String("sourceLanguage", "en-us", "")
+	outdir := mergeCmd.String("outdir", ".", "")
+	format := mergeCmd.String("format", "json", "")
+
+	mergeCmd.Parse(arguments)
+
+	if mergeCmd.Parsed() {
```
nicksnyder: this will always be true

goi18n/constants_command.go
```diff
@@ -86,6 +95,56 @@ func (cc *constantsCommand) execute() error {
 	return nil
 }
 
+func (cc *constantsCommand) parse(arguments []string) {
+	constantsCmd := flag.NewFlagSet("constants", flag.ExitOnError)
```
nicksnyder: `flags` would probably be a better variable name here

goi18n/merge_command.go
```diff
@@ -78,6 +81,30 @@ func (mc *mergeCommand) execute() error {
 	return nil
 }
 
+func (mc *mergeCommand) parse(arguments []string) {
+	mergeCmd := flag.NewFlagSet("merge", flag.ExitOnError)
```
nicksnyder: `flags`

goi18n/constants_command.go
```diff
@@ -0,0 +1,237 @@
...
+	switch v := reflect.ValueOf(t); v.Kind() {
+	case reflect.Map:
+		for _, k := range []language.Plural{"zero", "one", "two", "few", "many", "other"} {
+			vt := v.MapIndex(reflect.ValueOf(k))
+			if !vt.IsValid() {
+				continue
+			}
+
+			if len(result) > 0 {
+				result += "\n// "
```
nicksnyder: It would be better if this sort of formatting were in the template itself.

For example

```
type templateConstants struct {
    ID      string
    Name    string
    Comments []string
}
```

Then in the template you can range over `Comments`.

goi18n/constants_command.go
```diff
@@ -34,6 +41,7 @@ var constTemplate = template.Must(template.New("").Parse(`// DON'T CHANGE THIS F
 
 package {{.PackageName}}
 {{range .Constants}}
+// {{.Name}} {{.Comment}}
```
rodrigocorsi2: this model will generate name and comment on different lines, but golint demands name and comment on the same line
```go
// Name Comment
```

nicksnyder: My suggestion should yield a comment that passes lint.

e.g.

```
// DDays is the identifier for the following localizable string template(s):
// one: "{{.Count}} day"
// other: "{{.Count}} days"
const DDays = "d_days"
```

rodrigocorsi2:
@nicksnyder done

PR #53, opened 2016-10-16
handle Count field from template data

mh-cbon:
```sh
$ go test ./... -bench=.
PASS
ok  	github.com/nicksnyder/go-i18n/goi18n	0.007s
PASS
ok  	github.com/nicksnyder/go-i18n/i18n	0.004s
BenchmarkTranslateNonPluralWithMap-4             	 1000000	      3343 ns/op
BenchmarkTranslateNonPluralWithStruct-4          	  500000	      4071 ns/op
BenchmarkTranslateNonPluralWithStructPointer-4   	  500000	      5344 ns/op
BenchmarkTranslatePluralWithMap-4                	 1000000	      4363 ns/op
BenchmarkTranslatePluralWithMapAndCountField-4   	  500000	      3845 ns/op
BenchmarkTranslatePluralWithStruct-4             	  300000	      5993 ns/op
BenchmarkTranslatePluralWithStructPointer-4      	  300000	      4641 ns/op
PASS
ok  	github.com/nicksnyder/go-i18n/i18n/bundle	17.785s
BenchmarkNewOperand-4   	^[[A 2000000	       818 ns/op
PASS
ok  	github.com/nicksnyder/go-i18n/i18n/language	2.497s
?   	github.com/nicksnyder/go-i18n/i18n/language/codegen	[no test files]
BenchmarkExecuteNilTemplate-4          	300000000	         3.75 ns/op
BenchmarkExecuteHelloWorldTemplate-4   	500000000	         3.74 ns/op
BenchmarkExecuteHelloNameTemplate-4    	 1000000	      1934 ns/op
BenchmarkSprintf-4                     	10000000	       397 ns/op
PASS
ok  	github.com/nicksnyder/go-i18n/i18n/translation	9.950s

$ go test ./...
ok  	github.com/nicksnyder/go-i18n/goi18n	0.016s
ok  	github.com/nicksnyder/go-i18n/i18n	0.004s
ok  	github.com/nicksnyder/go-i18n/i18n/bundle	0.004s
ok  	github.com/nicksnyder/go-i18n/i18n/language	0.014s
?   	github.com/nicksnyder/go-i18n/i18n/language/codegen	[no test files]
ok  	github.com/nicksnyder/go-i18n/i18n/translation	0.008s
```

One q i have, now there s a `strconv.Atoi` if Count field value is a string.

Which may return an error, see [this](https://github.com/mh-cbon/go-i18n/blob/master/i18n/bundle/bundle.go#L269)

I m unsure if it should panic, for instance it just ignores the parameter, thus, the template rendering operation will probably fails, which, if i m right, should display the translation id, without any warning.

The "without a warning" bother me, but panicking bother me too.

:- /

### DIFF
--- a/i18n/bundle/bundle.go
+++ b/i18n/bundle/bundle.go
@@ -260,6 +260,11 @@ func (b *Bundle) translate(lang *language.Language, translationID string, args .
 			dataMap["Count"] = count
 			data = dataMap
 		}
+	} else {
+		dataMap := toMap(data)
+		if c, ok := dataMap["Count"]; ok {
+			count = c
+		}
 	}
 
 	p, _ := lang.Plural(count)
--- a/i18n/example_test.go
+++ b/i18n/example_test.go
@@ -30,6 +30,15 @@ func Example() {
 	fmt.Println(T("person_unread_email_count", 1, bobStruct))
 	fmt.Println(T("person_unread_email_count", 2, bobStruct))
 
+	type Count struct{ Count int }
+	fmt.Println(T("your_unread_email_count", Count{0}))
+	fmt.Println(T("your_unread_email_count", Count{1}))
+	fmt.Println(T("your_unread_email_count", Count{2}))
+
+	fmt.Println(T("your_unread_email_count", map[string]interface{}{"Count": 0}))
+	fmt.Println(T("your_unread_email_count", map[string]interface{}{"Count": "1"}))
+	fmt.Println(T("your_unread_email_count", map[string]interface{}{"Count": "3.14"}))
+
 	fmt.Println(T("person_unread_email_count_timeframe", 3, map[string]interface{}{
 		"Person":    "Bob",
 		"Timeframe": T("d_days", 0),
@@ -39,6 +48,22 @@ func Example() {
 		"Timeframe": T("d_days", 1),
 	}))
 	fmt.Println(T("person_unread_email_count_timeframe", 3, map[string]interface{}{
+		"Person":    "Bob",
+		"Timeframe": T("d_days", 2),
+	}))
+
+	fmt.Println(T("person_unread_email_count_timeframe", 1, map[string]interface{}{
+		"Count":     30,
+		"Person":    "Bob",
+		"Timeframe": T("d_days", 0),
+	}))
+	fmt.Println(T("person_unread_email_count_timeframe", 2, map[string]interface{}{
+		"Count":     20,
+		"Person":    "Bob",
+		"Timeframe": T("d_days", 1),
+	}))
+	fmt.Println(T("person_unread_email_count_timeframe", 3, map[string]interface{}{
+		"Count":     10,
 		"Person":    "Bob",
 		"Timeframe": T("d_days", 2),
 	}))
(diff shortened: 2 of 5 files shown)


i18n/bundle/bundle.go
```diff
@@ -260,6 +261,22 @@ func (b *Bundle) translate(lang *language.Language, translationID string, args .
 			dataMap["Count"] = count
 			data = dataMap
 		}
+	} else {
+		dataMap := toMap(data)
+		if c, ok := dataMap["Count"]; ok {
+			if cStr, ok := c.(string); ok {
+				var err error
+				count, err = strconv.Atoi(cStr)
```
nicksnyder: Why is this more complicated than the original PR (#49)?

You don't need to convert the string to a number because the library already handles this conversion later on this line:

```go
p, _ := lang.Plural(count)
```

i18n/example_test.go
```diff
@@ -30,6 +30,15 @@ func Example() {
...
 	fmt.Println(T("person_unread_email_count", 2, bobStruct))
 
+	type Count struct{ Count int }
+	fmt.Println(T("your_unread_email_count", Count{0}))
+	fmt.Println(T("your_unread_email_count", Count{1}))
+	fmt.Println(T("your_unread_email_count", Count{2}))
+
+	fmt.Println(T("your_unread_email_count", map[string]interface{}{"Count": 0}))
+	fmt.Println(T("your_unread_email_count", map[string]interface{}{"Count": "1"}))
+	fmt.Println(T("your_unread_email_count", map[string]interface{}{"Count": "3.14"}))
```
nicksnyder: Can you also add tests for "person_unread_email_count_timeframe" to show that it works with other template data?

It would also be nice to have a test for what happens if both are provided (a count parameter and a template with a Count variable).

i18n/i18n.go
```diff
@@ -69,9 +69,15 @@ import (
 // If translationID is a non-plural form, then the first variadic argument may be a map[string]interface{}
 // or struct that contains template data.
 //
-// If translationID is a plural form, then the first variadic argument must be an integer type
+// If translationID is a plural form, the function accepts two parameter signatures
+// - T(count int, data struct{})
```
nicksnyder: I think this would be slightly more readable as a numbered list (i.e. `1.` here instead of `-`).

nicksnyder:
The library is designed not to panic, but if consumers of the library wish to panic or do some error handling, then they can wrap the T function and detect when the returned translation is equal to the translation id.

i18n/example_test.go
mh-cbon: np, both are added.

mh-cbon:
ok thanks, anyway the changes removed that err, so the question does not matter anymore

i18n/bundle/bundle.go
```diff
@@ -261,6 +261,12 @@ func (b *Bundle) translate(lang *language.Language, translationID string, args .
 			data = dataMap
 		}
 	}
+	{
```
nicksnyder: This should probably be an `else` right?

```go
if count != nil {
    // current logic
} else {
    dataMap := toMap(data)
    if c, ok := dataMap["Count"]; ok {
        count = c
    }
}
```

mh-cbon:
we got an else ! BTW, i just noticed the travis file may need a little update  for 1.6.

Issue #61, opened 2017-01-01
Add TOML support


bep:
As discussed in this thread:

https://github.com/spf13/hugo/issues/2577

We use go-i18n in Hugo and it works great, but the number one related question we get on the support forum comes from people having a hard time getting their language files in YAML right (indentation issues, mostly).

So it would be nice if go-i18n could support the simpler TOML format and also optionally support a slightly flatter data format, so we could write:

```TOML
[[d_days]]
one = "{{.Count}} day"
other = "{{.Count}} days"

[[string_id_2]]
other = "Some other translation"
``` 

/cc @moorereason @bogem

nicksnyder:
I am supportive of this change and would like to see this done in two separate pull requests:
1. Add support for flat format (while maintaining support for current format) using existing serializations (JSON, YAML).
2. Add support for TOML.

I _may_ have a chance to work on this in 2-3 weeks, but can't promise anything. If someone else wants to drive this please volunteer.

moorereason:
Split flatter data format enhancement off as Issue #62.

PR #65, opened 2017-02-20
Add flatter data file structure

### LINKED ISSUE #62: Add flatter data file structure
A flatter data file structure would be beneficial.  Implementation should maintain support of existing format.

### JSON
```json
{
  "d_days": {
    "one": "{{.Count}} day",
    "other": "{{.Count}} days"
  },
  "string_id_2": {
    "other": "Some other translation"
  }
}
```

### YAML
```yaml
d_days:
  one: "{{.Count}} day"
  other: "{{.Count}} days"

string_id_2:
  other: Some other translation
```

### TOML (assuming #61)
```toml
[d_days]
one = "{{.Count}} day"
other = "{{.Count}} days"

[string_id_2]
other = "Some other translation"
```

bogem:
Fixes #62
Updates #61 

TODO:
- [x] Add docs 

Please leave your opinions and comments.

/cc @bep @moorereason @nicksnyder


i18n/bundle/bundle.go
```diff
@@ -78,34 +78,79 @@ func (b *Bundle) ParseTranslationFileBytes(filename string, buf []byte) error {
 }
 
 func parseTranslations(filename string, buf []byte) ([]translation.Translation, error) {
+	// if it's standard format
+	var standardFormat []map[string]interface{}
+	err := unmarshal(filename, buf, &standardFormat)
+	if err == nil {
+		translations, err := parseStandardFormat(standardFormat)
```
nicksnyder: this can just be

```
return parseStandardFormat(standardFormat)
```

i18n/bundle/bundle.go
```diff
@@ -78,34 +78,79 @@ func (b *Bundle) ParseTranslationFileBytes(filename string, buf []byte) error {
...
+		}
+		return translations, nil
+	}
+
+	// if it's flatter format
+	var flatterFormat map[string]map[string]interface{}
+	if err := unmarshal(filename, buf, &flatterFormat); err != nil {
+		return nil, err
+	}
+	translations, err := parseFlatterFormat(flatterFormat)
```
nicksnyder: `return parseFlatterFormat(flatterFormat)`

i18n/bundle/bundle.go
```diff
@@ -78,34 +78,79 @@ func (b *Bundle) ParseTranslationFileBytes(filename string, buf []byte) error {
...
 		if err != nil {
-			return nil, fmt.Errorf("unable to parse translation #%d in %s because %s\n%v", i, filename, err, translationData)
+			return nil, fmt.Errorf("unable to parse translation #%d because %s\n%v", i, err, translationData)
 		}
 		translations = append(translations, t)
 	}
 	return translations, nil
 }
 
+func parseFlatterFormat(data map[string]map[string]interface{}) ([]translation.Translation, error) {
```
nicksnyder: let's just call this `parseFlatFormat` (and update files to be `.flat.json`)

i18n/bundle/bundle.go
```diff
@@ -78,34 +78,79 @@ func (b *Bundle) ParseTranslationFileBytes(filename string, buf []byte) error {
...
-		return nil, fmt.Errorf("unsupported file extension %s", format)
+		return fmt.Errorf("unsupported file extension %s", format)
 	}
 
-	var translationsData []map[string]interface{}
 	if len(buf) > 0 {
-		if err := unmarshalFunc(buf, &translationsData); err != nil {
-			return nil, fmt.Errorf("failed to load %s because %s", filename, err)
-		}
+		return unmarshalFunc(buf, out)
```
nicksnyder: can you add back the error message? It is nice to know which file the error happened in.

```
if err := unmarshalFunc(buf, &translationsData); err != nil {
	return fmt.Errorf("failed to load %s because %s", filename, err)
}
```

i18n/translations_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func testFile(t *testing.T, path string) {
+	b := bundle.New()
+	b.MustLoadTranslationFile(path)
+
+	T, err := b.Tfunc("en-US")
+	if err != nil {
+		panic(err)
```
nicksnyder: no reason to panic in a test, just fail the test

i18n/translations_test.go
```diff
@@ -0,0 +1,107 @@
...
+	id   string
+	arg  interface{}
+	want string
+}{}
+
+func init() {
+	initTestCases()
+}
+
+func initTestCases() {
```
nicksnyder: no need to create a separate method, just do this in `init`

i18n/translations_test.go
```diff
@@ -0,0 +1,107 @@
...
+	b := bundle.New()
```
nicksnyder: it doesn't appear necessary to create a bundle during initialization

i18n/translations_test.go
```diff
@@ -0,0 +1,107 @@
...
+	b.MustLoadTranslationFile(path)
+
+	T, err := b.Tfunc("en-US")
+	if err != nil {
+		panic(err)
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.want, func(t *testing.T) {
```
nicksnyder: using `Run` means tests don't compile on Go < 1.7. Can we do without?

goi18n/testdata/en-us.flatter.json
```diff
@@ -0,0 +1,34 @@
+{
```
nicksnyder: if this is a copy of `en-us.all.json`, can we call this `en-us.all.flat.json`?

goi18n/testdata/en-us.flatter.yaml
```diff
@@ -0,0 +1,25 @@
+program_greeting:
```
nicksnyder: `en-us.all.flat.yaml`

i18n/translations_test.go
bogem: It is used for creating a Tfunc and Tfunc is used for checking "d_days" in `init` (person_unread_email_count_timeframe).

README.md
```diff
@@ -118,6 +118,49 @@ Here is an example of the default file format that go-i18n supports:
 
 To use a different file format, write a parser for the format and add the parsed translations using [AddTranslation](https://godoc.org/github.com/nicksnyder/go-i18n/i18n#AddTranslation).
 
+Flat Format
```
nicksnyder: I think we need to update the `goi18n merge` command to have an option to output the flat format before we advertise this on the main readme. Up to you whether you want to just remove the README update from this PR or update the merge command.

i18n/translations_test.go
nicksnyder: Ah, I see. This is weird because you are reading the JSON file during init but testing other formats in the actual tests.

I would actually not try to do anything in init. Since it is just a test, it is fine to re-init the small array of test cases for each test.

moorereason:
I'm not familiar with the internals of the JSON and YAML packages.  Do they return early if the data structure doesn't match the target interface?  Can the caller avoid a double-unmarshal when using the flat format?

README.md
bogem: How do you offer to implement this? I propose to add a new boolean flag called `-flat` to output data in flat format

bogem:
@moorereason reading the API of `encoding/json` and `gopkg.in/yaml.v2` I didn't find any method to check the data structure, but in json lib there is a variety of errors (`SyntaxError`, `UnmarshalFieldError`, `UnmarshalTypeError` and etc.), so we can know, if it's flat format or a user just mistook in json syntax (forgot semicolon, colon, curly bracket and etc.)

bep:
This change was motivated by another format, `TOML`, so having special casings for `JSON` is sub-optimal. 

I think the current implementation is great as is.

If there is a simple way to detect the format ala:

```go
isFlat := bytes.Contains(buf[:100], []byte("someflatidentified")
```

Then that would be great, but I would not add it if it adds complexity.

bogem:
The standard format could be detected this way:
isStandard := bytes.Contains(data[:100], []byte("id")) && (bytes.Contains(data[:100], []byte("translation")) || bytes.Contains(data[:100], []byte("translations")))

The problem could be if id will be more than ~95 bytes length.

bogem:
@bep TOML doesn't support standard format layout. That's why we should add flat format
If user passes TOML translation file, it will be automatically parsed as flat format. So we don't need to detect the format for TOML. At least I will implement like that

bogem:
I think, the simplest and right way to detect the format is built on the first byte.
**If** it's JSON and first byte of data is `[` **or** if it's YAML and first byte is `-`, **then** it's standard format, **else** it's flat format.

nicksnyder:
I know TOML isn't part of this PR (and it shouldn't be), but what would the equivalent TOML file look like?

> I think, the simplest and right way to detect the format is built on the first byte.
If it's JSON and first byte of data is [ or if it's YAML and first byte is -, then it's standard format, else it's flat format.

This sounds right to me.

bogem:
@moorereason 
> @bep TOML doesn't support standard format layout. That's why we should add flat format
> If user passes TOML translation file, it will be automatically parsed as flat format. So we don't need to detect the format for TOML. At least I will implement like that

README.md
bogem: Done

bogem:
So I updated merge command and made tests for it. What do you think about it?

i18n/bundle/bundle.go
```diff
@@ -78,34 +78,89 @@ func (b *Bundle) ParseTranslationFileBytes(filename string, buf []byte) error {
...
+}
+
+// unmarshal finds an appropriate unmarshal function for ext
+// (extension of filename) and unmarshals buf to out. out must be a pointer.
+func unmarshal(ext string, buf []byte, out interface{}) error {
+	var err error
+	switch ext {
 	case ".json":
-		unmarshalFunc = json.Unmarshal
+		err = json.Unmarshal(buf, out)
```
nicksnyder: just `return json.Unmarshal(buf, out)`

i18n/bundle/bundle.go
```diff
@@ -78,34 +78,89 @@ func (b *Bundle) ParseTranslationFileBytes(filename string, buf []byte) error {
...
 	case ".yaml":
-		unmarshalFunc = yaml.Unmarshal
+		err = yaml.Unmarshal(buf, out)
```
nicksnyder: same

goi18n/merge_command.go
```diff
@@ -88,67 +83,81 @@ func (mc *mergeCommand) parse(arguments []string) {
...
 }
 
-type marshalFunc func(interface{}) ([]byte, error)
-
-func (mc *mergeCommand) writeFile(label string, translations []translation.Translation, localeID string, marshal marshalFunc) error {
+func (mc *mergeCommand) writeFile(label string, translations []translation.Translation, localeID string) error {
 	sort.Sort(translation.SortableByID(translations))
-	buf, err := marshal(marshalInterface(translations))
+
+	var iTranslations interface{}
```
nicksnyder: not a fan of this variable name. I would just avoid it though

```
var marshal func ([]translation.Translation) interface{}
if mc.flat {
    marshal = marshalFlatInterface
} else {
    marshal = marshalInterface
}

buf, err := mc.marshal(marshal(translations))
```

bogem:
@nicksnyder I've already told you: TOML doesn't support standard format layout. That's why we should add flat format. If user passes TOML translation file, it will be automatically parsed as flat format. So we don't need to detect the format for TOML. At least I will implement like that

goi18n/merge_command.go
bogem: `mc.marshal(marshal(translations))` - it's also very strange.
I offer to name it as `conv2I` or just `convert`.

nicksnyder:
I know TOML doesn't support standard format. I want to know what it will
look like in the flat format.

bogem:
@nicksnyder 
```go
func isStandardFormat(ext string, buf []byte) bool {
  if ext == ".toml" {
    return false
  }

  firstRune := rune(buf[0])
  if (ext == ".json" && firstRune == '[') || (ext == ".yaml" && firstRune == '-') {
    return true
  }
  return false
}
```

nicksnyder:
Sorry if I wasn't clear. I want to know what `en-us.flat.toml` or perhaps just `en-us.toml` will look like.

bogem:
Oh sorry. I thought, you asked me about detection of TOML 😅
`en-us.flat.toml` will look like:
```toml
[program_greeting]
other = "Hello world"

[person_greeting]
other = "Hello {{.Person}}"

[my_height_in_meters]
one = "I am {{.Count}} meter tall."
other = "I am {{.Count}} meters tall."

[your_unread_email_count]
one = "You have {{.Count}} unread email."
other = "You have {{.Count}} unread emails."

[person_unread_email_count]
one = "{{.Person}} has {{.Count}} unread email."
other = "{{.Person}} has {{.Count}} unread emails."

[person_unread_email_count_timeframe]
one = "{{.Person}} has {{.Count}} unread email in the past {{.Timeframe}}."
other = "{{.Person}} has {{.Count}} unread emails in the past {{.Timeframe}}."

[d_days]
one = "{{.Count}} day."
other = "{{.Count}} days."
```

PR #67, opened 2017-04-02
bundle: Delete leading comments and whitespaces before detecting a format

bogem:


bep:
What if this is how I start my YAML file?

```yaml
# Some comment
- id: translations
  translation: 
    one: "Translation"
    other: "Translations"
 ```

bogem:
Then it should be also deleted. Same for JSON

nicksnyder:
@bogem can you update to trim comments too?

bogem:
Yes, I'm in process

PR #73, opened 2017-06-06
Allow use custom language codes

### LINKED ISSUE #72: go-i18n fails on custom language codes
If user wants to have custom language code that aren't in [Unicode CLDR](http://www.unicode.org/cldr/charts/28/supplemental/language_plural_rules.html) (like at.yaml, dk.yaml, ch.yaml, gb.yaml, etc.), `bundle.ParseTranslationFileBytes` returns `no language found in "dk.yaml"`.
I don't think that users should be limited only with CLDR languages. They can freely use language codes they want and it can be technically implemented.
Discussion: https://discuss.gohugo.io/t/i18n-country-codes-as-part-of-the-url/6881

bogem:
Fix #72 
Fix https://github.com/spf13/hugo/issues/3564

go-i18n doesn't return an error now, if user provides non-registered language code.

/cc @bep @nicksnyder

### DIFF
--- a/i18n/language/language.go
+++ b/i18n/language/language.go
@@ -17,6 +17,15 @@ type Language struct {
 	*PluralSpec
 }
 
+// NewLanguage returns Language with specified tag and pluralSpec.
+// If pluralSpec is nil, it uses EmptyPluralSpec.
+func NewLanguage(tag string, pluralSpec *PluralSpec) *Language {
+	if pluralSpec == nil {
+		return &Language{Tag: tag, PluralSpec: EmptyPluralSpec}
+	}
+	return &Language{Tag: tag, PluralSpec: pluralSpec}
+}
+
 func (l *Language) String() string {
 	return l.Tag
 }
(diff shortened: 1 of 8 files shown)


i18n/language/pluralspec.go
```diff
@@ -38,7 +45,6 @@ func (ps *PluralSpec) Plural(number interface{}) (Plural, error) {
 // getPluralSpec returns the PluralSpec that matches the longest prefix of tag.
 // It returns nil if no PluralSpec matches tag.
 func getPluralSpec(tag string) *PluralSpec {
-	tag = NormalizeTag(tag)
```
nicksnyder: Why did you decide to move this out of the implementation here? It looks like it is always normalized before passing in so this change seems to just bloat the diff.

bogem: I did it to get rid of a double normalizing in language.Parse.

i18n/language/language.go
```diff
@@ -39,41 +48,65 @@ func (l *Language) MatchingTags() []string {
...
 		}
 	}
-	return deduped
+	return unique
+}
+
+// ParseFirst returns Language with first found language code in src.
+// It also supports custom, non-provided in Unicode CLDR language codes.
+// It can parse language tags and Accept-Language headers.
+func ParseFirst(src string) *Language {
```
nicksnyder: Why do we need a new method for this? Why can't you update `Parse` to do what you need?

i18n/language/language.go
```diff
@@ -39,41 +48,65 @@ func (l *Language) MatchingTags() []string {
 // Parse returns a slice of supported languages found in src or nil if none are found.
 // It can parse language tags and Accept-Language headers.
 func Parse(src string) []*Language {
```
nicksnyder: This function is now completely unused, so I would want to remove it, but that would be an unnecessary backward incompatible change. It seems like `ParseFirst` can/should just call `Parse` and get the first element returned.

bogem: >  It seems like ParseFirst can/should just call Parse and get the first element returned.

I will better do this way.

bogem: > It seems like ParseFirst can/should just call Parse and get the first element returned.

No, `Parse` returns ** only registered** languages, so we need a separate function to parse a language regardless of either it's Unicode CLDR language code or custom language code.

> This function is now completely unused, so I would want to remove it, but that would be an unnecessary backward incompatible change.

You're right. That's why I stayed it there.

i18n/language/language.go
bogem: > Why can't you update Parse to do what you need?

Just because of backward compatibility. I don't want to break the behaviour of `Parse` function, because maybe someone already uses it.

bogem:
> I am not convinced that this is better than just exporting registerPluralSpec and requiring each project to explicitly register fake locales. Is there a reason you don't like that approach?

We should export it anyway if user wants to have plural specs. But in Hugo we should develop a simple way to add a possibility for user to add custom plural specs and it needs a lot of time actually. That's why at first I just allow user to have files with custom language tags, but later I think we will export registerPluralSpec.

bep:
@bogem as long as there is a way to detect if a language code is "valid go-toml", then Hugo can simply register fake plural specs for all non-valids.

bogem:
@bep It's a good idea, but subjectively I think it's not a native to process custom language tags.

bep:
@bogem what do you mean by "not a native to process custom language tags"?

bogem:
I mean it's not natural and not an elegant solution.

bogem:
Ping

nicksnyder:
> as long as there is a way to detect if a language code is "valid go-toml", then Hugo can simply register fake plural specs for all non-valids. The "user defined plural spec" is pretty far outside of this scope.

I agree with this.

> But in Hugo we should develop a simple way to add a possibility for user to add custom plural specs and it needs a lot of time actually.

I am happy to implement features that benefit Hugo, however the design of said features in go-i18n need to be motivated by first principals, and not necessarily what is best for Hugo.

As is, here are my concerns with the PR.
1) It leaves a completely unused function `Parse`. If it is unused then it should be deleted, and we need a major version bump. I don't think this change is worth a major version bump.
2) It adds a brand new function `ParseFirst` which seems to duplicate the functionality of `Parse` except with a minor difference and doesn't reuse the implementation of `Parse`.
3) It is not obvious to me that go-i18n should just blindly provide some default behavior for unrecognized tags. Being explicit about registering language tags you expect to use seems very reasonable.

If we think that go-i18n should provide default behavior (I personally don't), then we should go all in and do a major version bump.

My counter proposal is to make a small character change to export [registerPluralSpec] (https://sourcegraph.com/github.com/nicksnyder/go-i18n/-/blob/i18n/language/pluralspec.go#L21:18). This will be a very small diff and does not require a major version bump. Hugo will need to do work to figure out what specs to register.

If that work turns out to be hard in some way, you can perhaps propose a new function like `registerDefaultPluralSpec`.

bep:
>My counter proposal is to make a small character change to export [registerPluralSpec] 

This would be more than sufficient if there is a way to detect if a plural spec already exists with a given id.

nicksnyder:
You can export [getPluralSpec](https://sourcegraph.com/github.com/nicksnyder/go-i18n/-/blob/i18n/language/pluralspec.go#L40:6-40:19) too.

bogem:
I close this PR and propose a change later easier to implement and discuss.

bogem:
> If that work turns out to be hard in some way, you can perhaps propose a new function like registerDefaultPluralSpec.

This function will break the behaviour of go-i18n, if user wants to parse files like `en-us.all.json`. `language.Parse("en-us.all.json")` returns `[en-us all json]` and I think, user expects that only first prefix will be parsed.

nicksnyder:
`registerDefaultPluralSpec` would not need to break any current behavior. Existing code would not be calling that method so it would behave as before. We would get to define what the new behavior is if you call that function.

bogem:
Yes, you're right, but I meant different thing:
I mean that `RegisterDefaultPluralSpec` doesn't make sense, because if user sets default plural spec and provides translation files, `language.Parse` will always return multiple languages (e.g. `language.Parse("en.json") = ["en", "json"]`) and `bundle.ParseTranslationFileBytes` will always return error about multiple languages on these files.

bogem:
Please take a look at #74.

PR #74, opened 2017-06-15
Export registerPluralSpec function

### LINKED ISSUE #72: go-i18n fails on custom language codes

bogem:
Fix #72 

The example of registering PluralSpec of unknown language:
```go
filename := "dk.yaml"
langs := language.Parse(filename)
if len(langs) == 0 {
    tags := language.AllLanguageTags(filename)
    if len(tags) == 0 {
        panic("no languages found in " + filename)
    }
    language.RegisterPluralSpec(tags[0], language.EmptyPluralSpec)
}
i18n.MustLoadTranslationFile(filename)
```

/cc @nicksnyder @bep

### DIFF
--- a/i18n/language/language_test.go
+++ b/i18n/language/language_test.go
@@ -7,8 +7,8 @@ import (
 
 func TestParse(t *testing.T) {
 	tests := []struct {
-		src  string
-		lang []*Language
+		src   string
+		langs []*Language
 	}{
 		{"en", []*Language{{"en", pluralSpecs["en"]}}},
 		{"en-US", []*Language{{"en-us", pluralSpecs["en"]}}},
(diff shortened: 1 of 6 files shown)


bogem:
Any comments?

i18n/language/language_test.go
```diff
@@ -62,13 +62,60 @@ func TestParse(t *testing.T) {
...
+		{"en.json", []string{"en"}},
+		{"en-US.json", []string{"en-us"}},
+		{"en-us.json", []string{"en-us"}},
+		{"en-xx.json", []string{"en-xx"}},
+		{"unknown", []string{"unknown"}},
+		{"", nil},
+	}
+	for _, test := range tests {
+		langs := AllLanguageTags(test.src)
+		if !areStringSlicesEqual(langs, test.expected) {
```
nicksnyder: is there something wrong with using reflect.DeepEqual?

bogem: It's just because I don't like `reflect` package 😅

nicksnyder: Less code is better. Reflect performance doesn't matter for testing.

bogem:
> Sorry for delay, I was on vacation. I would have insta-merged this PR if it were simply exporting the RegisterPluralSpec function as discussed, but it includes more than that.
My thinking was that if RegisterPluralSpec is exported, then non-standard languages can be registered before Parse is called (so there should not be any need to change the Parse method or add a new function).

It's not so simple as we thought.

When I exported RegisterPluralSpec, I ran into some problems. I will try to clearly explain them.
```git
-type operands struct {
+// Operand is used for determing the number and its properties.
+// More: http://unicode.org/reports/tr35/tr35-numbers.html#Operands
+type Operand struct {

@nicksnyder
Why is this change necessary?
```
If user wants to register PluralSpec, he/she obviously have to create some PluralSpec. 
Previous specification of PluralSpec:
```go
type PluralSpec struct {
	Plurals    map[Plural]struct{}
	PluralFunc func(*operands) Plural
}
```
The `operands` is private type. So how user can declare the PluralFunc, if the argument is of the private type? The obvious solution I found is to export `operands`.

i18n/language/operands.go
```diff
@@ -6,8 +6,9 @@ import (
 	"strings"
 )
 
-// http://unicode.org/reports/tr35/tr35-numbers.html#Operands
-type operands struct {
+// Operand is used for determing the number and its properties.
+// More: http://unicode.org/reports/tr35/tr35-numbers.html#Operands
+type Operand struct {
```
bogem: Explanation here: https://github.com/nicksnyder/go-i18n/pull/74#issuecomment-311573341

bogem:
> I don't understand what the utility of AllLanguageTags is.

`AllLanguageTags` returns all string tags found in src. `Parse` returns only known language
As you can see in example in first comment, `AllLanguageTags` is very useful for automated registering of unknown languages.

bogem:
Ping

nicksnyder:
I merged https://github.com/nicksnyder/go-i18n/pull/79 which I think should be sufficient for your needs, but if it isn't we can continue the discussion on https://github.com/nicksnyder/go-i18n/issues/72

PR #75, opened 2017-06-17
language: Save MatchingTags

bogem:
Small performance improvement
```
name                   old time/op    new time/op    delta
LoadTransitionFiles-4     354µs ± 1%     351µs ± 1%  -0.87%  (p=0.001 n=15+15)

name                   old alloc/op   new alloc/op   delta
LoadTransitionFiles-4    54.6kB ± 0%    54.5kB ± 0%  -0.26%  (p=0.000 n=14+14)

name                   old allocs/op  new allocs/op  delta
LoadTransitionFiles-4       942 ± 0%       933 ± 0%  -0.96%  (p=0.000 n=15+15)
```


nicksnyder:
This change is backwards incompatible and it looks like the perf gain isn't significant, so I am going to leave this PR open until there are other backward incompatible changes being made.

bogem:
Yes, it successfully compiles.
``` go
$ cat lang.go
package main

import "github.com/nicksnyder/go-i18n/i18n/language"

func main() {
	_ = language.Language{"ru", language.EmptyPluralSpec}
}

$ go run lang.go

```

i18n/language/language.go
```diff
@@ -17,23 +19,35 @@ type Language struct {
 	*PluralSpec
 }
 
+// NewLanguage returns new Language with specified tag and spec.
+func NewLanguage(tag string, spec *PluralSpec) *Language {
```
nicksnyder: If this is a backward compatible change, why is this new method necessary or beneficial? I would rather just leave it out.

bogem: I made it only because it's a good tone that you should use a constructor for creating objects, not using the `&Language{}` annotation.

nicksnyder: I don't see the benefit of having a factory method at this point. The factory method doesn't do anything special so I would rather not add it unless it becomes necessary.

bogem: Actually, there is a problem when we want to create `&Language{}` inside the package.
E.g. having `&Language{tag, spec}` compiler complains: 
```
i18n/language/language.go:58: cannot use tag (type string) as type []string in field value
i18n/language/language.go:58: cannot use spec (type *PluralSpec) as type string in field value
```
`NewLanguage` solves it.

nicksnyder: The call sites are not specifying field names. You can either explicitly specify field names 
 like this `&Language{Tag: tag, PluralSpec: spec}` or move the declaration of `matchingTags` to the bottom of the struct.

```
type Language struct {
  	// Tag uniquely identifies the language as defined by RFC 5646.
  	//
  	// Most language tags are a two character language code (ISO 639-1)
 	// optionally followed by a dash and a two character country code (ISO 3166-1).
 	// (e.g. en, pt-br)
 	Tag string

  	*PluralSpec

	matchingTags []string
  }
```

I think the latter would be ideal because it is more backwards compatible. Also, please add doc for `matchingTags`.

bogem: Using latter construction compiler returns:
```
i18n/language/language.go:60: too few values in struct initializer
i18n/language/language.go:68: too few values in struct initializer
i18n/language/language.go:74: too few values in struct initializer
...
```

nicksnyder: Is this only a problem inside of the package? If these types of errors happen outside of the package then this isn't a backwards compatible change.

bogem: Yes, it's only inside the package.

bogem: No, it's a problem outside of a package too.

bogem: That's why I created NewLanguage func.

nicksnyder: If this is a problem outside of the package, then this whole PR is backwards incompatible.

bogem: You're right

PR #86, opened 2018-02-02
Make tests pass with go1.10

jcajka:
avoiding following tests failures:

i18n/translations_test.go:66: Error call has possible formatting directive %v
FAIL	_/root/go-i18n/i18n [build failed]
i18n/language/pluralspec_test.go:73: Errorf format %q has arg spec of wrong type *language.PluralSpec
FAIL	_/root/go-i18n/i18n/language [build failed]

more details in Go1.10 upstream release notes https://tip.golang.org/doc/go1.10#test

### DIFF
--- a/i18n/language/pluralspec_test.go
+++ b/i18n/language/pluralspec_test.go
@@ -70,7 +70,7 @@ func TestGetPluralSpec(t *testing.T) {
 	for _, test := range tests {
 		spec := GetPluralSpec(test.src)
 		if spec != test.spec {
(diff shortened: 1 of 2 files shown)


nicksnyder:
Can we change the args to keep using %q?

jcajka:
I think it depends on what is expected output, but in general they are not integers. What are you thinking about? I can definitively change it that way.

nicksnyder:
`test.src` is a string and I would like it to be %q so the string is quoted

PR #92, opened 2018-04-05
V2 beta

nicksnyder:
Work in progress


bep:
This looks really, really good. I assume that old bundle files will still work after this?

nicksnyder:
> This looks really, really good. I assume that old bundle files will still work after this?

No, I have not spent any time thinking about backcompat yet; I designed this fresh.

I think the best way to deal with old message files is to run a one-time conversion via `goi18n`. Thoughts?

mh-cbon:
>  How to version goi18n command?

might this help https://github.com/golang/dep/issues/221

bep:
>I think there should be a really good reason to break the file formats.

Note that I'm happy to do the work if it's a really good reason to, i.e. the change itself adds really useful stuff. But not just for vanity reasons ("this looks prettier").

nicksnyder:
@bep the runtime library can load message files from v1, and message files from v1 will automatically get reformated to the new structure the first time `goi18n merge` is run.

Issue #99, opened 2018-04-27
Ability to add custom text/template functions


Seklfreak:
I think it could be very useful to be able to add custom functions to the internally used template engine.

The Template method that allows it is https://golang.org/pkg/text/template/#Template.Funcs .

I'm not sure where it would the best place would be to implement this. Maybe adding another Field `Funcs` of the type `template.funcMap` to the `Message` struct would be a good idea?

Seklfreak:
When trying to work on a possible implementation I realised `LocalizeConfig` should be the better place. I made a PR #100 .

PR #100, opened 2018-04-27
adds ability to define custom funcs in LocalizeConfig #99

Seklfreak:
This PR adds the ability to define custom template engine funcs. It adds a `Funcs` field to the `LocalizeConfig` struct and passes this parameter to `gotemplate.Funcs`.

### DIFF
--- a/v2/internal/template_test.go
+++ b/v2/internal/template_test.go
@@ -7,7 +7,7 @@ import (
 
 func TestParse(t *testing.T) {
 	tmpl := &Template{Src: "hello"}
-	if err := tmpl.parse("", ""); err != nil {
+	if err := tmpl.parse("", "", nil); err != nil {
 		t.Fatal(err)
 	}
 	if tmpl.ParseErr == nil {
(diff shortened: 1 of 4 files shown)


v2/internal/template_test.go
```diff
@@ -21,7 +22,25 @@ func TestParse(t *testing.T) {
...
-	if err := tmpl.parse("", ""); err != expectedErr {
+	if err := tmpl.parse("", "", nil); err != expectedErr {
 		t.Fatalf("expected %#v; got %#v", expectedErr, err)
 	}
 }
+
+func TestParseWithFunc(t *testing.T) {
+	tmpl := &Template{Src: "hello"}
+	funcs := template.FuncMap{
+		"foo": func() string {
```
nicksnyder: This test doesn't seem to actually use the "foo" func. Can you add a test that does that?

Seklfreak:
forgot to change `Src`, done! :)

v2/internal/template_test.go
```diff
@@ -21,7 +22,25 @@ func TestParse(t *testing.T) {
...
+	if err := tmpl.parse("", "", nil); err != expectedErr {
 		t.Fatalf("expected %#v; got %#v", expectedErr, err)
 	}
 }
+
+func TestParseWithFunc(t *testing.T) {
+	tmpl := &Template{Src: "{{foo}}"}
+	funcs := template.FuncMap{
+		"foo": func() string {
+			return "bar"
```
nicksnyder: Can you also check that the value produced by the template is actually “bar”

PR #113, opened 2018-07-17
Update plural for language Vietnamese

thuandoan274:
Hope you soon approve my pull request for my project progress

### DIFF
--- a/i18n/language/codegen/plurals.xml
+++ b/i18n/language/codegen/plurals.xml
@@ -12,7 +12,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 
         <!-- 1: other -->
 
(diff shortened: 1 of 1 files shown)


nicksnyder:
The xml file shouldn’t be updated directly. You can follow the process here:
https://github.com/nicksnyder/go-i18n/blob/master/dev.md

If you think the data is wrong, you should file an issue and/or submit a pull request on the [upstream repo](https://unicode.org/cldr/trac/newticket).

Thanks!

thuandoan274:
Maybe I misunderstanding about plural. I think I can use plural count for 2 difference sentences for just 1 key
```[CommentOnAReview]
one = "{{.Name}} just commented your review"
other = "{{.Name}} and {{.PluralCount}} people has commented your review"
```
But It's not working with Vi, maybe I should split them to be 2 difference keys

Issue #119, opened 2018-08-03
V2 - get list of available translations from Bundle


Trane9991:
Hello, I would to be able to get the list of all available translations, something like:
```
	bundle := &i18n.Bundle{DefaultLanguage: language.English}
	bundle.RegisterUnmarshalFunc("toml", toml.Unmarshal)
	// No need to load active.en.toml since we are providing default translations.
	// bundle.MustLoadMessageFile("active.en.toml")
	bundle.MustLoadMessageFile("active.es.toml")

      for _, tr := range  bundle.ListTranslations() {
        ...
     }
```

PR #120, opened 2018-08-03
add func to list Language tags from bundle

Trane9991:
Fixes the #119


Trane9991:
@nicksnyder is something more need from me here?

PR #122, opened 2018-08-13
Fixed localizer example

beono:
Type for `TemplateData` should be `map[string]interface`, because the value for `Count` is integer


nicksnyder:
Thanks!

PR #124, opened 2018-09-21
Export MessageNotFoundErr

bep:
So a client can react to that particular error situation.

### DIFF
--- a/v2/i18n/localizer.go
+++ b/v2/i18n/localizer.go
@@ -74,11 +74,12 @@ func (e *invalidPluralCountErr) Error() string {
 	return fmt.Sprintf("invalid plural count %#v for message id %q: %s", e.pluralCount, e.messageID, e.err)
 }
 
-type messageNotFoundErr struct {
+// MessageNotFoundErr is returned from Localize when a messate could not be found.
+type MessageNotFoundErr struct {
 	messageID string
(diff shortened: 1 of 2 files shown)


bep:
Note that this comes from #125 -- which I guess could replace this PR if that worked. But I think it would be cleaner to just check the error.

Also, I don't understand the "codecov error". You may want to add something ala this: https://github.com/bep/s3deploy/blob/master/codecov.yml

nicksnyder:
I don't have a problem with this PR but it would be helpful to me (for my own understanding) if you could explain your use-case for explicitly wanting to check for this error.

bep:
If a message with a given ID in a given language does not exist, I want to be able to both 

1. Print a WARNING that there is a missing translation
2. Print the default language's version as a fallback

nicksnyder:
Do you still think this is useful to export?

bep:
>Do you still think this is useful to export?

Yes, as the API is now, these are the cases:

1. Found in the current language
2. Not found in the current language, but in some other language (check language tag)
3. Not found in any language (check error type)
4. Another error ... fail.

The API above could maybe be improved, but I cannot see how 3) can be handled without a way to check the error type (something ala `os.IsNotExist`).

Issue #125, opened 2018-09-21
Odd behaviour with DefaultMessage


bep:
Working on https://github.com/gohugoio/hugo/pull/5243 and trying to find a workaround for #124 I found some odd behaviour with `DefaultMessage`
```go
translated, err := localizer.Localize(&i18n.LocalizeConfig{
				//DefaultMessage: defaultMessage,
				MessageID:    translationID,
				TemplateData: templateData,
			})
```

I have a test case which returns `translated` as expected and no error with the code above. If I uncomment the `DefaultMessage`, I always get the default message in return.

nicksnyder:
I don't understand the issue you are reporting and what you expected to happen. Can you clarify?

bep:
* My intuition says that `DefaultMessage` is the message returned when no match can be found in the bundle with `translationID`. This way I can use that to determine if a string is considered untranslated.
* What I'm seeing instead is that if `DefaultMessage` is provided, that is what I always get.

bep:
This is approximately my failing case:

```go
func TestLocalizer_DefaultMessage(t *testing.T) {
	en := "[hello]\nother = \"Hello, World!\""
	es := "[hello]\nother = \"¡Hola, Mundo!\""
	bndl := &Bundle{DefaultLanguage: language.English}
	bndl.RegisterUnmarshalFunc("toml", toml.Unmarshal)

	_, err := bndl.ParseMessageFileBytes([]byte(en), "en.toml")
	if err != nil {
		t.Fatal(err)
	}
	_, err = bndl.ParseMessageFileBytes([]byte(es), "es.toml")
	if err != nil {
		t.Fatal(err)
	}

	var defaultMessage = &Message{
		ID:    "___I18N_DEFAULT",
		Other: "I18N_MISSING",
	}

	_ = defaultMessage

	localizer := NewLocalizer(bndl, "es")

	translated, err := localizer.Localize(&LocalizeConfig{
		DefaultMessage: defaultMessage,
		MessageID:      "hello",
	})

	if err != nil {
		t.Fatal(err)
	}

	if translated != "¡Hola, Mundo!" {
		t.Fatalf("Got %s, expected %s", translated, "¡Hola, Mundo!")
	}

}
```

If fails with:

```bash
 Got I18N_MISSING, expected ¡Hola, Mundo!
```

This test, however, passes:

```go
func TestLocalizer_DefaultMessage(t *testing.T) {
	en := "[hello]\nother = \"Hello, World!\""
[...]

PR #130, opened 2018-09-24
Add failing test for DefaultMessage

bep:
See #125


bep:
I will have to look closer into this.

PR #144, opened 2018-11-24
add .golangci.yml

jirfag:
This pull request fixes warning on fetching dependencies in [golangci.com](https://golangci.com/r/github.com/nicksnyder/go-i18n) analysis

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,5 @@
+# https://github.com/golangci/golangci/wiki/Configuration
+
+service:
+  prepare:
+    - go get -t ./...


nicksnyder:
Thanks for reaching out! As you saw, I was also experimenting with `go mod download` in https://github.com/nicksnyder/go-i18n/pull/145. Is there  any reason to prefer one over the other?

jirfag:
I prefer `go get` because it works. `go mod` doesn't work as you see in [this output](https://github.com/nicksnyder/go-i18n/pull/145#issuecomment-441418938) - I will try to investigate it and suggest solution with go modules tomorrow

Issue #135, opened 2018-10-04
Nested .yml file


maeglindeveloper:
Hi everyone, 

I'm actually trying to use the `MustLocalize(...)` method to load message defined in a .yml language file.
I have a doubt  regarding the ability to load nested yml messages as follow

**en.yml** file:
```
mysection:
  title: my title
  content: my content
```
using that main function:

```
func main() {
	bundle := &i18n.Bundle{DefaultLanguage: language.English}
	bundle.RegisterUnmarshalFunc("yml", yaml.Unmarshal)
	bundle.MustLoadMessageFile("en.yml")

	localizer := i18n.NewLocalizer(bundle, "en")

	title := localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "mysection.title"})
	content := localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "mysection.content"})

        ...
```
It fails saying that **mysection.title** is not found.

The only way I found is to have something like that:

**en.yml** file:
```
mysection_title: my title
mysection_content: my content
```
and 
```
	title := localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "mysection_title"})
	content := localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "mysection_content"})
```

Maybe I did it wrong :s.
Any thoughts ?
Thanks in advance
@nicksnyder

nicksnyder:
A translation file is just keys and values. No nesting is supported.

You can use comments or you can use separate files to group your translations.

See also https://github.com/nicksnyder/go-i18n/issues/121

PR #157, opened 2019-03-29
Nested translation support

### LINKED ISSUE #121: Can get text from nested object ?
I wanna organize my file text to be more clarification like this
```
[Tip]
    NotEnoughCoin = "Not enough coin"

[Follow]
    HasAFollowing = "Has someone just following you"
```
Thank you!

benoitmasson:
Fixes #121 and #135 

This PR introduces support for nested messages.

For example (using JSON syntax):
* `{"nested": {"hello": "world"}}` defines a message with ID `"nested.hello"` and value `"world"`
* `{"nested": {"hello": "world", "foo": "bar"}}` defines a message with ID `"nested.hello"` and value `"world"` and a message with ID `"nested.foo"` and value `"bar"`
* `{"outer": {"nested": {"inner": "value"}}}` defines a message with ID `"outer.nested.inner"` and value `"value"`

The following previous behaviors are preserved:
* `{"notnested.hello": "world"}}` defines a message with ID `"notnested.hello"` and value `"world"`
* `{"notnested": {"other": "world"}}` defines a message with ID `"notnested"` and value `"world"`
* `{"notnested": {"description": "world", "dummy": "nothing"}}` defines a message with ID `"notnested"` and value `"other"`

(in other words, as soon as a key is found in the reserved set `"id"`, `"description"`, `"hash"`, `"leftdelim"`, `"rightdelim"`, `"zero"`, `"one"`, `"two"`, `"few"`, `"many"`, `"other"` with a meaningful value, it is considered not nested).

So, to include a reserved key as message ID, use nesting:
* `{"nested": {"description": {"other": "world"}}}` defines a message with ID `"nested.description"` and value `"world"`

Unit tests were added to enforce both behaviors and non-regression.
Manually tested using YAML and JSON translation files.

Note: there is probably a performance loss at parsing time due to that update, since maps are scanned twice, instead of once (a first time to select whether it is a map with reserved keys, a second time to build the messages).

### DIFF
--- a/v2/internal/message.go
+++ b/v2/internal/message.go
@@ -2,8 +2,6 @@ package internal
 
 import (
 	"fmt"
-	"reflect"
-	"sort"
 	"strings"
 )
 
@@ -47,15 +45,6 @@ type Message struct {
 	// justCreated is used internally, to know whether the message ID
 	// has to be set (creation) or prepended with prefix (recursive call)
 	justCreated bool
-}
-
-// Equal compares a message to another, ignoring private fields.
-func (m *Message) Equal(m2 *Message) bool {
-	justCreated := m.justCreated   // backup
-	m.justCreated = m2.justCreated // copy value, to ignore it on testing deep equality
-	eq := reflect.DeepEqual(m, m2)
-	m.justCreated = justCreated // restore
-	return eq
 }
 
 // NewMessage parses data and returns a new message.
@@ -222,27 +211,4 @@ func isMessage(v interface{}) bool {
 		}
 	}
 	return false
-}
-
-// EqualMessages compares two slices of messages, ignoring private fields and order.
-// Sorts both input slices, which are therefore modified by this function.
-func EqualMessages(m1, m2 []*Message) bool {
-	if len(m1) != len(m2) {
-		return false
-	}
-
-	var less = func(m []*Message) func(int, int) bool {
-		return func(i, j int) bool {
-			return m[i].ID < m[j].ID
-		}
-	}
-	sort.Slice(m1, less(m1))
-	sort.Slice(m2, less(m2))
-
-	for i, m := range m1 {
-		if !m.Equal(m2[i]) {
-			return false
-		}
-	}
-	return true
 }
--- a/v2/internal/message.go
+++ b/v2/internal/message.go
@@ -41,10 +41,6 @@ type Message struct {
 
 	// Other is the content of the message for the CLDR plural form "other".
 	Other string
-
-	// justCreated is used internally, to know whether the message ID
-	// has to be set (creation) or prepended with prefix (recursive call)
-	justCreated bool
 }
 
 // NewMessage parses data and returns a new message.
@@ -71,7 +67,6 @@ func (m *Message) unmarshalInterface(v interface{}) error {
 	if err != nil {
 		return err
 	}
-	m.justCreated = true
 	for k, v := range strdata {
 		switch strings.ToLower(k) {
 		case "id":
--- a/v2/internal/message.go
+++ b/v2/internal/message.go
@@ -41,6 +41,10 @@ type Message struct {
 
 	// Other is the content of the message for the CLDR plural form "other".
 	Other string
+
+	// justCreated is used internally, to know whether the message ID
+	// has to be set (creation) or prepended with prefix (recursive call)
+	justCreated bool
 }
 
 // NewMessage parses data and returns a new message.
@@ -67,6 +71,7 @@ func (m *Message) unmarshalInterface(v interface{}) error {
 	if err != nil {
 		return err
 	}
+	m.justCreated = true
 	for k, v := range strdata {
 		switch strings.ToLower(k) {
 		case "id":
@@ -161,4 +166,49 @@ func stringMap(v interface{}) (map[string]string, error) {
 	default:
 		return nil, fmt.Errorf("unsupported type %#v", value)
 	}
+}
+
+// isMessage tells whether the given data is a message, or a map containing
+// nested messages.
+// A map is assumed to be a message if it contains any of the "reserved" keys:
+// "id", "description", "hash", "leftdelim", "rightdelim", "zero", "one", "two", "few", "many", "other"
+// with a string value.
+// e.g.,
+// - {"message": {"description": "world"}} is a message
+// - {"message": {"description": "world", "foo": "bar"}} is a message ("foo" key is ignored)
+// - {"notmessage": {"description": {"hello": "world"}}} is not
+// - {"notmessage": {"foo": "bar"}} is not
+func isMessage(v interface{}) bool {
+	reservedKeys := []string{"id", "description", "hash", "leftdelim", "rightdelim", "zero", "one", "two", "few", "many", "other"}
+	switch data := v.(type) {
+	case string:
+		return true
+	case map[string]interface{}:
+		for _, key := range reservedKeys {
+			val, ok := data[key]
+			if !ok {
+				continue
+			}
+			_, ok = val.(string)
+			if !ok {
+				continue
+			}
+			// v is a message if it contains a "reserved" key holding a string value
+			return true
+		}
+	case map[interface{}]interface{}:
+		for _, key := range reservedKeys {
+			val, ok := data[key]
+			if !ok {
+				continue
+			}
+			_, ok = val.(string)
+			if !ok {
+				continue
+			}
+			// v is a message if it contains a "reserved" key holding a string value
+			return true
+		}
+	}
+	return false
 }
--- a/v2/internal/message.go
+++ b/v2/internal/message.go
@@ -2,6 +2,8 @@ package internal
 
 import (
 	"fmt"
+	"reflect"
+	"sort"
 	"strings"
 )
 
@@ -45,6 +47,15 @@ type Message struct {
 	// justCreated is used internally, to know whether the message ID
 	// has to be set (creation) or prepended with prefix (recursive call)
 	justCreated bool
+}
+
+// Equal compares a message to another, ignoring private fields.
+func (m *Message) Equal(m2 *Message) bool {
+	justCreated := m.justCreated   // backup
+	m.justCreated = m2.justCreated // copy value, to ignore it on testing deep equality
+	eq := reflect.DeepEqual(m, m2)
+	m.justCreated = justCreated // restore
+	return eq
 }
 
 // NewMessage parses data and returns a new message.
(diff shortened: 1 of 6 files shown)


v2/internal/parse.go
```diff
@@ -39,49 +40,115 @@ func ParseMessageFileBytes(buf []byte, path string, unmarshalFuncs map[string]Un
...
+		messages = make([]*Message, 0, len(data))
 		for _, data := range data {
-			m, err := NewMessage(data)
+			// recursively scan slice items
+			childMessages, err := recGetMessages(data, false)
 			if err != nil {
 				return nil, err
 			}
-			messageFile.Messages = append(messageFile.Messages, m)
+			for _, m := range childMessages {
```
golangcibot: S1011: should replace loop with `messages = append(messages, childMessages...)` (from `gosimple`)

v2/internal/message.go
```diff
@@ -41,6 +43,19 @@ type Message struct {
 
 	// Other is the content of the message for the CLDR plural form "other".
 	Other string
+
+	// justCreated is used internally, to know whether the message ID
+	// has to be set (creation) or prepended with prefix (recursive call)
+	justCreated bool
```
nicksnyder: Is there any way to solve this problem without needing to add an unexported field to this struct?

benoitmasson: Did it! I agree, this was not very smart… I managed to use the `isMessage()` function instead for the same purpose.

This one is costly, so I anticipated the call (and pass the result as a parameter to `recGetMessages`) to avoid computing it twice for each submap.

While I was at it, I also factorized some duplicate code, this will make these functions easier to read and maintain, and at the same time improving coverage (reducing the content of the untested `switch` branches to trivial code).

v2/internal/parse_test.go
```diff
@@ -28,23 +33,180 @@ func TestParseMessageFileBytes(t *testing.T) {
...
 		actual, err := ParseMessageFileBytes([]byte(testCase.file), testCase.path, testCase.unmarshalFuncs)
-		if err != testCase.err {
-			t.Fatalf("expected error %#v; got %#v", testCase.err, err)
+		if (err == nil && testCase.err != nil) ||
+			(err != nil && testCase.err == nil) ||
+			(err != nil && testCase.err != nil && err.Error() != testCase.err.Error()) {
+			t.Errorf("%s failed: expected error %#v; got %#v", testCase.name, testCase.err, err)
+			continue
+		}
+		if actual == nil {
```
nicksnyder: I don't think this condition serves any purpose and therefore should be deleted.

v2/internal/parse_test.go
```diff
@@ -28,23 +33,180 @@ func TestParseMessageFileBytes(t *testing.T) {
...
+			continue
 		}
 		if actual.Path != testCase.messageFile.Path {
-			t.Fatalf("expected path %q; got %q", testCase.messageFile.Path, actual.Path)
+			t.Errorf("%s failed: expected path %q; got %q", testCase.name, testCase.messageFile.Path, actual.Path)
```
nicksnyder: Instead of prepending the test name to each error, it would be worth it to use t.Run for subtests.

Examples: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/nicksnyder/go-i18n%24+t.Run

v2/internal/parse_test.go
```diff
@@ -28,23 +33,180 @@ func TestParseMessageFileBytes(t *testing.T) {
...
+			continue
 		}
-		if !reflect.DeepEqual(actual.Messages, testCase.messageFile.Messages) {
-			t.Fatalf("expected %#v; got %#v", testCase.messageFile.Messages, actual.Messages)
+	}
+}
+
+// equalMessages compares two slices of messages, ignoring private fields and order.
+// Sorts both input slices, which are therefore modified by this function.
+func equalMessages(m1, m2 []*Message) bool {
```
nicksnyder: I don't think this function is necessary anymore?

v2/internal/parse.go
```diff
@@ -39,49 +40,103 @@ func ParseMessageFileBytes(buf []byte, path string, unmarshalFuncs map[string]Un
...
+	isChildMessage := isMessage(data)
+	childMessages, err := recGetMessages(data, isChildMessage, false)
+	if err != nil {
+		return nil, err
+	}
+	for _, m := range childMessages {
+		if isChildMessage {
+			if m.ID == "" {
+				m.ID = id // start with innermost key
+			}
```
nicksnyder: Can you explain to me when the else case happens here? I am just trying to follow the logic.

v2/internal/parse_test.go
benoitmasson: Actually yes, it is useful: when the `ParseMessageFileBytes()` call raises an error, `actual` is nil and the remaining lines of the test panic.

Would you prefer that I write `if err != nil` for the test instead? It is strictly equivalent, maybe more explicit. Another option would be to write the previous test as follows:
```go
if err != nil || testCase.err != nil {
	if err == nil || testCase.err == nil || err.Error() != testCase.err.Error()) {
		t.Errorf("%s failed: expected error %#v; got %#v", testCase.name, testCase.err, err)
	}
	continue
}
```
which forces to `continue` the for loop when an error is found, and makes the `actual` test useless.

What do you like best?

v2/internal/parse_test.go
benoitmasson: It is, for the `"multiple nested test"` subtest: this test returns 2 messages nested under the same key.
Since map keys enumeration is random in Go, in theory these 2 messages can be created in any order (in practice, first line is created first 90% of the time though), which makes the test fail randomly if we do not make order-insensitive comparison.

It is required only in `TestParseMessageFileBytes()`, so I moved it in `parse_test.go` as an unexported function, but we need it.

v2/internal/parse.go
benoitmasson: Indeed, I did not get this line correct from the start, it is not so obvious :-)

All messages created by `recGetMessages()` use `NewMessage()` to setup message content and fieds. When `NewMessage()` returns, it sets the message fields it finds inside the message submap: if `id` was defined, then it is set, otherwise it is not and we need to set it manually with the map key.

In practice, this is tested with 2 tests:
* "basic test" (`{"hello": "world"}`) creates a message with `Other` content `"world"`, but no `ID`, so we manually set it to `"hello"` when iterating through the high-level map (line 132).
* "basic test with id" (`{"key": {"id": "forced.id"}}`) creates a message with no content, but with `ID: "forced.id"`. In this case, we do not want to reset the `ID` to `"key"` (passes test line 130, but not the one line 131).

v2/internal/parse_test.go
```diff
@@ -28,23 +33,180 @@ func TestParseMessageFileBytes(t *testing.T) {
...
+					ID:    "prepended.hello",
+					Other: "world",
+				}},
+			},
+		},
+		{
+			name: "invalid test (no key)",
+			file: `"hello"`,
+			path: "en.json",
+			err:  errors.New("invalid translation file, expected key-values, got a single value"),
```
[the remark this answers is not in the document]
benoitmasson: Done, commit `87b9968`

v2/internal/parse_test.go
benoitmasson: I hesitated when I introduced the `name` attribute, but did not want to change test structure ;-)
I also prefer to use subtests, fixed in commit https://github.com/nicksnyder/go-i18n/pull/157/commits/f0d608c5543d2c116ed88c48b7839158da3b5737

v2/internal/parse_test.go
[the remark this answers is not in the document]
nicksnyder: I was actually wondering about saving the error itself instead of just the message because I think that would simplify the code in the tests that checks for the equality of errors.

e.g.

```go

var errInvalidTranslationFile = errors.New("invalid translation file, expected key-values, got a single value")
```

v2/internal/parse_test.go
nicksnyder: Ah, I see what you mean. I think I would find it clearer written like this:

```go
if testCase.messageFile == nil && actual != nil || testCase.messageFile != nil && actual == nil {
    t.Fatalf("expected message file %#v; got %#v", testCase.messageFile, actual)
}

if testCase.messageFile != nil {
    if actual.Path != testCase.messageFile.Path {
        t.Fatalf("expected path %q; got %q", testCase.messageFile.Path, actual.Path)
    }
    if actual.Tag != testCase.messageFile.Tag {
        t.Fatalf("expected tag %q; got %q", testCase.messageFile.Tag, actual.Tag)
    }
    if actual.Format != testCase.messageFile.Format {
        t.Fatalf("expected format %q; got %q", testCase.messageFile.Format, actual.Format)
    }
    if !equalMessages(actual.Messages, testCase.messageFile.Messages) {
        t.Fatalf("expected %#v; got %#v", testCase.messageFile.Messages, actual.Messages)
    }
}
```

v2/internal/parse_test.go
```diff
@@ -28,23 +33,182 @@ func TestParseMessageFileBytes(t *testing.T) {
...
-			t.Fatalf("expected tag %q; got %q", testCase.messageFile.Tag, actual.Tag)
-		}
-		if actual.Format != testCase.messageFile.Format {
-			t.Fatalf("expected format %q; got %q", testCase.messageFile.Format, actual.Format)
+		t.Run(testCase.name, func(t *testing.T) {
+			actual, err := ParseMessageFileBytes([]byte(testCase.file), testCase.path, testCase.unmarshalFuncs)
+			if (err == nil && testCase.err != nil) ||
+				(err != nil && testCase.err == nil) ||
+				(err != nil && testCase.err != nil && err.Error() != testCase.err.Error()) {
+				t.Errorf("expected error %#v; got %#v", testCase.err, err)
```
nicksnyder: Might as well use `t.Fatalf` and omit the return.

v2/internal/parse_test.go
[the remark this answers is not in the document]
benoitmasson: no problem, replaced commit 87b9968 with https://github.com/nicksnyder/go-i18n/commit/6d4c4423af15735809ce0e755a158228b56f4d21

v2/internal/parse_test.go
benoitmasson: Done, commit beea385

v2/internal/parse_test.go
benoitmasson: Indeed, I also changed that in commit beea385.

I kept the last outputs as `Errorf()` to allow to continue with next assertion: only the error and messageFile tests are fatal and prevent to continue.

PR #158, opened 2019-04-12
Fall back to an 'other' plural lookup before giving up

gabrieljackson:
This change alters the v1 behavior so that plural lookups where
the plural count is greater than 1, but no translation is found,
will attempt another translation with 'other' before falling back
to the default language. This will only happen if the first
translation attempt was not 'other' to begin with.

The reasoning here is that this behavior should be preferable
in most instances instead of resorting to a different language
immediately.

Edit: if this sort of behavior is desirable to merge then I will
address the failing tests.

### DIFF
--- a/i18n/bundle/bundle.go
+++ b/i18n/bundle/bundle.go
@@ -380,7 +380,16 @@ func (b *Bundle) translate(lang *language.Language, translationID string, args .
 	p, _ := lang.Plural(count)
 	template := translation.Template(p)
 	if template == nil {
-		return translationID
+		if p == language.Other {
+			return translationID
+		}
+		countInt, ok := count.(int)
+		if ok && countInt > 1 {
+			template = translation.Template(language.Other)
+			if template == nil {
+				return translationID
+			}
+		}
 	}
 
 	s := template.Execute(data)


nicksnyder:
This seems like a reasonable best effort behavior, but I am slightly worried about changing behavior at this point. It is possible that users of v1 depend on this function returning the translation ID to detect/report when translations are missing.

gabrieljackson:
I understand and agree that although this behavior may be desirable in more cases than not, it could affect users expecting the ID in v1.

Do you have any thoughts on how we may be able to get this behavior merged here in a reasonable way? That would obviously be desirable over having a fork just for this change.

I will be honest that I haven't had time to look into v2 much yet and what it would take for us to move over to it. I am not sure if migrating to v2 would possibly give us the ability to add this fallback behavior.

nicksnyder:
Can you explain what motivated you to propose this change?

Why not just copy the "other" translations to the appropriate plural rules in the translation file?

v2 explicitly returns an error in this case:
https://github.com/nicksnyder/go-i18n/blob/master/v2/internal/message_template.go#L55-L58

gabrieljackson:
The motivation for this comes from the process with which we manage and create our various translation files. A template is uploaded to our Pootle server and then various translation files are created over time as the community provides additional translations.

There are times when translations are missing and being able to fall back to an "other" translation, which is almost always available, would provide a better experience for us than defaulting to another language immediately.

nicksnyder:
I'll implement this in v2, but am not interested in changing v1 behavior at this point.

See https://github.com/nicksnyder/go-i18n/pull/174

PR #175, opened 2019-05-09
update to CLDR v35.1 (2019-04-17)

nicksnyder:
The build is broken because the rule for `kw` is inconsistent with the examples. Specifically the rule categorizes "1000.0", "10000.0" and "100000.0" as `two` but these are listed as examples for `other`.

```
--- FAIL: TestKw (0.00s)
    rule_test.go:30: kw: PluralFormFunc(&plural.Operands{N:1000, I:1000, V:1, W:0, F:0, T:0}) returned "two", <nil>; expected "other"
    rule_test.go:30: kw: PluralFormFunc(&plural.Operands{N:10000, I:10000, V:1, W:0, F:0, T:0}) returned "two", <nil>; expected "other"
    rule_test.go:30: kw: PluralFormFunc(&plural.Operands{N:100000, I:100000, V:1, W:0, F:0, T:0}) returned "two", <nil>; expected "other"
```

### DIFF
--- a/v2/internal/plural/codegen/plurals.xml
+++ b/v2/internal/plural/codegen/plurals.xml
@@ -6,7 +6,7 @@ CLDR data files are interpreted according to the LDML specification (http://unic
 For terms of use, see http://www.unicode.org/copyright.html
 -->
 <supplementalData>
-    <version number="$Revision: 14397 $"/>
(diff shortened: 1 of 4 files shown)


nicksnyder:
Upstream bug: https://github.com/unicode-org/cldr/pull/6

PR #178, opened 2019-05-23
remove v2 path

HeChuanXUPT:
update import path


nicksnyder:
The import path still includes v2 with Go modules

PR #179, opened 2019-05-24
remove v2

HeChuanXUPT:


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/nicksnyder/go-i18n/v2
+module github.com/nicksnyder/go-i18n
```
nicksnyder: I can’t merge this because the v2 suffix is necessary to avoid breaking backward compatibility with v1. Can you explain the problem you have that motivated this PR?

HeChuanXUPT: I got some errors when go build my program

nicksnyder: What are you using to vendor? What version of Go are you using?

HeChuanXUPT: go version 1.10

nicksnyder:
And what are you using to generate your vendor folder?

HeChuanXUPT:
I use go vendor to manage packages
> mkdir -p vendor/github.com/nicksnyder/
> cd vendor/github.com/nicksnyder/
> git clone https://github.com/nicksnyder/go-i18n.git

go.mod
nicksnyder: Can you upgrade to 1.10.3+? I think that might solve your problem.

nicksnyder: More info here: https://github.com/golang/go/wiki/Modules#how-are-v2-modules-treated-in-a-build-if-modules-support-is-not-enabled-how-does-minimal-module-compatibility-work-in-197-1103-and-111

nicksnyder:
go-i18n v2 is configured to work with Go modules as documented here: https://github.com/golang/go/wiki/Modules

Note that Go modules works best with 1.10.3+, so the easiest fix is probably to upgrade your version of go to at least 1.11.

If helpful, you can checkout the v2.0.0 tag when the v2 code was in a v2 subfolder.
You could also fix your vendor folder to put the i18n package under a `v2` directory. If you upgrade to go 1.11+ then `go mod vendor` should do this for you.

I am aware that some of the go doc links are broken, and am waiting for a resolution here: https://github.com/golang/gddo/issues/567#issuecomment-496091172

I am going to close this PR because ultimately this is something that you will need to fix on your end, but I hope the above suggestions are helpful and sorry for the inconvenience.

HeChuanXUPT:
ok, thanks

nicksnyder:
I restored the v2 subdirectory for greater compatibility so you should be able to update to 2.0.2

PR #202, opened 2019-09-29
fix(goi18n): could not extract constant message id

ksegun:
If you specified a constant for message id e.g

```
package main

import "github.com/nicksnyder/go-i18n/v2/i18n"

const constID = "ConstantID"
			
var m = &i18n.Message{
    ID: constID,
    Other: "ID is a constant",
}
```

goi18n extract failed with `failed to marshal en strings to toml: toml: Key '' is not a valid table name. Key names cannot be empty.`

### DIFF
--- a/v2/goi18n/extract_command_test.go
+++ b/v2/goi18n/extract_command_test.go
@@ -186,11 +186,6 @@ zero = "Zero translation"
 			`,
 		},
 		{
-			name:     "id constant",
-			fileName: "file.go",
-			file:     `package main`,
-		},
-		{
 			name:     "global declaration",
 			fileName: "file.go",
 			file: `package main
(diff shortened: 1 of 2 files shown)


v2/goi18n/extract_command_test.go
```diff
@@ -185,6 +185,28 @@ zero = "Zero translation"
 			}
 			`,
 		},
+		{
```
nicksnyder: this appears to duplicate the first test in this list

ozankasikci:
Having this same problem. Please release a new version including this fix.

nicksnyder:
[2.0.3](https://github.com/nicksnyder/go-i18n/releases/tag/v2.0.3) is released

PR #208, opened 2020-01-03
fixes nil reference exception if localization text is empty, fixes #207

### LINKED ISSUE #207: nil reference exception if localization text is empty
### Reproduction steps:
have a localization file, e.g. active.en.toml, with the content `Test = ""` and run the following command in your terminal in the appropriate directory:
```bash
goi18n merge active.*.toml
```
to generate your updated `translate.*.toml` files.

### Analysis
The execution of the above mentioned command will cause a nil reference exception here
https://github.com/nicksnyder/go-i18n/blob/26334ab1db070c2801ba26550d7d6238f6c9da10/v2/goi18n/merge_command.go#L132
because a nil value was added here
https://github.com/nicksnyder/go-i18n/blob/26334ab1db070c2801ba26550d7d6238f6c9da10/v2/goi18n/merge_command.go#L128

MJacred:
command will now abort with explanatory error message printing the ID with an empty localization text

fixes #207


v2/goi18n/merge_command.go
```diff
@@ -125,12 +125,16 @@ func merge(messageFiles map[string][]byte, sourceLanguageTag language.Tag, outdi
 		}
 		templates := map[string]*i18n.MessageTemplate{}
 		for _, m := range mf.Messages {
-			templates[m.ID] = i18n.NewMessageTemplate(m)
+			template := i18n.NewMessageTemplate(m)
+			if template == nil {
+				return nil, fmt.Errorf("missing translation for message id \"%s\"", m.ID)
```
nicksnyder:
```suggestion
				return nil, fmt.Errorf("missing translation for message id %q", m.ID)
```

v2/goi18n/merge_command.go
```diff
@@ -125,12 +125,16 @@ func merge(messageFiles map[string][]byte, sourceLanguageTag language.Tag, outdi
...
+			}
+			templates[m.ID] = template
 		}
 		if mf.Tag == sourceLanguageTag {
 			for _, template := range templates {
 				if sourceMessageTemplates[template.ID] != nil {
-					return nil, fmt.Errorf("multiple source translations for id %s", template.ID)
+					return nil, fmt.Errorf("multiple source translations for id \"%s\"", template.ID)
```
nicksnyder:
```suggestion
					return nil, fmt.Errorf("multiple source translations for id %q", template.ID)
```

MJacred: > Thanks for filing an issue and proposing a fix! This diff proposes treating the missing translation as an error but it isn’t obvious to me that it should be. My instinct would be to just skip over the missing one.

I had the same incline to simply continue. But I also wanted to report the missing translation. My initial implementation was like this:
```golang
template := i18n.NewMessageTemplate(m)
if template == nil {
	fmt.Fprintf(os.Stderr, "missing translation for message id %q\n", m.ID)
	continue
}
templates[m.ID] = template
```

Shall I change the commit to the above mentioned code? I'll update the unittest change accordingly.

P.S.: thanks for `%q`, didn't know that one yet

v2/goi18n/merge_command_test.go
```diff
@@ -35,6 +35,50 @@ func TestMerge(t *testing.T) {
...
+		{
+			name:           "single identity, one localization text after missing one",
+			sourceLanguage: language.AmericanEnglish,
+			inFiles: map[string][]byte{
+				"one.en-US.toml": []byte(`
+1HelloMessage = ""
+Body = "Finally some text!"
+`),
+			},
+			outFiles:         map[string][]byte{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			outFiles: map[string][]byte{
```

v2/goi18n/main.go
```diff
@@ -153,3 +153,38 @@ func (lt languageTag) Tag() language.Tag {
...
 }
+
+const (
+	// Lean prints key-value-pairs in active language source file with table
+	Lean int8 = 1 << iota
+	// Full means the hash will be printed additionally to `Lean` to the active language source file as well
+	Full
+)
+
+type pretty struct{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
type pretty struct {
```

MJacred:
@nicksnyder 
> Thanks for filing an issue and proposing a fix! This diff proposes treating the missing translation as an error but it isn’t obvious to me that it should be. My instinct would be to just skip over the missing one.

everything is fixed, can you merge the pull request, please?

v2/goi18n/merge_command_test.go
```diff
@@ -35,6 +35,50 @@ func TestMerge(t *testing.T) {
 				"active.en-US.toml": []byte("1HelloMessage = \"Hello\"\n"),
 			},
 		},
+		{
```
nicksnyder: Can you reduce this to two tests total (one from this hunk of the diff and also the plural one from the last hunk)? The order of translations doesn't matter so the extra test cases don't really do anything.

v2/goi18n/merge_command.go
```diff
@@ -125,12 +125,17 @@ func merge(messageFiles map[string][]byte, sourceLanguageTag language.Tag, outdi
 		}
 		templates := map[string]*i18n.MessageTemplate{}
 		for _, m := range mf.Messages {
-			templates[m.ID] = i18n.NewMessageTemplate(m)
+			template := i18n.NewMessageTemplate(m)
+			if template == nil {
+				fmt.Fprintf(os.Stderr, "missing translation for message id %q\n", m.ID)
```
nicksnyder: It isn't obvious to me that this should log an error. Can we just remove this?

MJacred: I disagree upon this point. A missing translation should not be ignored silently. The minimum I could agree on would be to change it to `os.Stdout`

nicksnyder:
```suggestion
```

PR #236, opened 2020-10-31
fix fallback behaviour

### LINKED ISSUE #234: MustLocalize panics even when DefaultMessage is specified
**Version:** 2.1.1
**Code:**
```go
const InternalServerError = "internalServerError"

func FallbackMsg(id string, msg string) *i18n.Message {
	return &i18n.Message{
		ID:    id,
		One:   msg,
		Many:  msg,
		Other: msg,
	}
}

localizer.MustLocalize(&i18n.LocalizeConfig{
	MessageID: InternalServerError,
	DefaultMessage: FallbackMsg(InternalServerError,
		"{{.Mention}} Internal server error occurred, please try again later."),
		TemplateData: map[string]interface{}{
			"Mention": "Adam",
		},
})
```
**Error:** message "internalServerError" not found in language "language tag"

In the older version (v2.0.3) everything works fine, something is wrong with the update.

Kichiyaki:
fixes #234

### DIFF
--- a/v2/i18n/localizer.go
+++ b/v2/i18n/localizer.go
@@ -194,7 +194,7 @@ func (l *Localizer) getMessageTemplate(id string, defaultMessage *Message) (lang
 	if defaultMessage == nil {
 		return language.Und, nil, &MessageNotFoundErr{tag: tag, messageID: id}
 	}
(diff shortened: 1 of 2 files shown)


nicksnyder:
This is working as intended. If a client asks for `en-US` and there is no valid fallback, it will return an error and the message in the default language. You may choose to ignore the error and display the default message if you wish.

PR #246, opened 2021-02-25
Add support for loading bundle from fs.FS

antonlindstrom:
This commit aims to add a method for loading bundles from fs.FS. The
change is done to be able to get a nice way to load bundles from an
embedded file system.

Since this can be done by implementing the exported API and using
ParseMessageFileBytes externally, this is not a change that is a must
have. Rather a nice shorthand for using the fs.FS.

This is how it can be used:

    package main

    import (
            "embed"

            "github.com/nicksnyder/go-i18n/v2/i18n"
    )

    //go:embed locale.*.json
    var LocaleFS embed.FS

    func main() {
            bundle := i18n.NewBundle(language.English)
            bundle.RegisterUnmarshalFunc("json", json.Unmarshal)

            bundle.LoadMessageFileFS(LocaleFS, "locale.sv.json")
    }

Thanks!

### DIFF
--- a/v2/i18n/bundlefs.go
+++ b/v2/i18n/bundlefs.go
@@ -1,0 +1,18 @@
+// +build go1.16
+
+package i18n
+
+import (
+	"io/fs"
+)
+
+// LoadMessageFileFS is like LoadMessageFile but instead of reading from the
+// hosts operating system's file system it reads from the fs file system.
+func (b *Bundle) LoadMessageFileFS(fsys fs.FS, path string) (*MessageFile, error) {
+	buf, err := fs.ReadFile(fsys, path)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


hachi8833:
I definitely want the feature❤️

nicksnyder:
Build checks were broken on master but [I just fixed that](https://github.com/nicksnyder/go-i18n/commit/5d2ec5fba990a664cbe711f626a9a448f894b793) so could you rebase on master?

I think this will fail the Go 1.9 build. Can you use build tags to make this function available in newer versions of go without breaking older versions of Go?

antonlindstrom:
Thank you so much for the good review! I completely missed the backwards compatibility. I've moved the method to it's own file and added a build tag, tested with both 1.16 and 1.15 (pre io/fs).

PR #247, opened 2021-02-25
Add Bengali (BN)

MjAbuz:
Language - Internationalization Bengali (BN) for reference please check here https://github.com/MjAbuz/hugo-book/pull/2

### DIFF
--- a/v2/example/active.bn.toml
+++ b/v2/example/active.bn.toml
@@ -1,0 +1,11 @@
+হ্যালোনাম = "Hello {{.Name}}"
+
+[MyUnreadEmails]
+description = "আমার কাছে উনরেড ইমেইলের সংখ্যা"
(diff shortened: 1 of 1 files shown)


MjAbuz:
@nicksnyder looking forward to add Bengali (BN)

v2/example/active.bn.toml
```diff
@@ -0,0 +1,11 @@
+হ্যালোনাম = "Hello {{.Name}}"
```
MjAbuz, line 1: We don't have caps and small letters , what to do about it?
Hello = হ্যালো
Name = নাম

MjAbuz (commented):
@nicksnyder  what do you advise?

nicksnyder (commented):
This is just an example file so I don’t see a reason to add an example for a new language. If something isn’t working well for a specific language then can you file an issue?

PR #270, opened 2022-03-03
doc(localizer): add specification about language tag detection

ColinBoisLydia:
I've been stumbling across a frustrating issue with the underlying tag detection from `x/text/language.Matcher`: the detection algorithm will return a tag value based on a given score who can be surprising. 

Because it is not a bug, I just updated the documentation of `Localizer` in order to give some hint for future user.

I did two go-playground to show this more specifically https://go.dev/play/p/TMoszzbeEZ2

Here is a link to a related Github issue in x/text/language: https://github.com/golang/go/issues/49176

### DIFF
--- a/v2/i18n/localizer.go
+++ b/v2/i18n/localizer.go
@@ -9,6 +9,15 @@ import (
 )
 
 // Localizer provides Localize and MustLocalize methods that return localized messages.
+// Localize and MustLocalize methods use a language.Tag matching algorithm based
+// on the best possible value. This algorithm may cause an unexpected language.Tag returned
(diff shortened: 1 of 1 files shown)


nicksnyder:
Thanks for digging in to this! Is there any way we could make the comment more specific about how the ordering in memory might impact the result (potentially even linking to the issues)?

ColinBoisLydia:
You're welcome! I added an example in order to be a little less nebulous along with a link to the Github issue and the Go code base which contains a tone of informations

Thank you for this pkg :-)

PR #284, opened 2022-10-26
IMPORTANT: Issue: concurrent map read and map write, Solution: Conver…

shantanubansal:
IMPORTANT: Issue: concurrent map read and map write, Solution: Converting Maps to Sync Maps

### DIFF
--- a/v2/i18n/bundle.go
+++ b/v2/i18n/bundle.go
@@ -3,6 +3,7 @@ package i18n
 import (
 	"fmt"
 	"io/ioutil"
+	"sync"
 
 	"github.com/nicksnyder/go-i18n/v2/internal/plural"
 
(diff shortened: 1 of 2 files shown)


prathabk (approved):

shantanubansal:
@nicksnyder can you please check this?
Do proposed changes looks good to you?

nicksnyder:
Bundle is documented to not be goroutine safe. It has been a while since I thought about this library, and I am not sure if the right decision is to force synchronization in the library here or let consumers handle necessary synchronization. This is the kind of change I would want to think more about before merging, and at the same time, making changes to this library is not a priority for me right now.

nicksnyder:
Going to close this PR as it has conflicts and would need to think through what a solution should be.

Issue #292, opened 2023-01-06
Easy wrapper around this library


amitavaghosh1:
Obviously the examples provided are way to descriptive for a real world use. With already so much code we have to write in go, I don't think, I would like to see 10 line implementations per localized string. 

Here is an example wrapper you can use . Please feel free to improve it. 

```go
package locales

import (
	"encoding/json"
	"fmt"
	"strings"

	"github.com/nicksnyder/go-i18n/v2/i18n"
	"golang.org/x/text/language"
)

type Localizer struct {
	bundle *i18n.Bundle
}

var localizer Localizer

func NewLocalizer(path string, languages ...string) Localizer {
	path = strings.TrimSuffix(path, "/")

	bundle := i18n.NewBundle(language.English)
	bundle.RegisterUnmarshalFunc("json", json.Unmarshal)

	for _, lang := range languages {
		bundle.MustLoadMessageFile(fmt.Sprintf("%s/translations/%s.json", path, lang))
	}

	return Localizer{bundle: bundle}
}

func (l Localizer) Get(lang string, id string) string {
	localizer := i18n.NewLocalizer(l.bundle, lang)

	cfg := &i18n.LocalizeConfig{
		DefaultMessage: &i18n.Message{
			ID:    id,
			Other: id,
			One:   id,
		},
	}

	str, err := localizer.Localize(cfg)
	if err != nil {
		return id
	}

	return str
}

func (l Localizer) GetWithData(lang, id string, data map[string]interface{}) string {
	localizer := i18n.NewLocalizer(l.bundle, lang)

	cfg := &i18n.LocalizeConfig{
		DefaultMessage: &i18n.Message{
			ID:    id,
			Other: id,
		},
		TemplateData: data,
	}
	str, err := localizer.Localize(cfg)
	if err != nil {
		return id
	}

	return str
}
```

and use it as such

```go
lc := locales.NewLocalizer(dotConfig.LocaleRootPath, "en", "es")

fmt.Println(lc.Get("en", "time_up"))
fmt.Println(lc.Get("es", "time_up"))
fmt.Println(lc.Get("es", "time up"))
```

translations/en.json
```json
{
    "time_up": "Retry limit exceeded"
}
```

translations/es.json
```json
{
    "time_up": "demasiados reintentos"
}
```

The immediate improvement is to make sure, the languages exists.

brutalzinn:
Thanks for this. Save me a lot of time. A small contribute: 
`
type Localizer struct {
[...]

PR #295, opened 2023-03-08
add feature change package name and message type name

mr-tron:
It can be useful for creating simple wrappers like https://github.com/nicksnyder/go-i18n/issues/292 asks. For example my wrapper:

```
package i18n

import (
	"embed"
	"github.com/BurntSushi/toml"
	"github.com/nicksnyder/go-i18n/v2/i18n"
	"golang.org/x/text/language"
)

//go:embed locale.*.toml
var localeFS embed.FS
var bundle *i18n.Bundle

func init() {
	bundle = i18n.NewBundle(language.English)
	bundle.RegisterUnmarshalFunc("toml", toml.Unmarshal)
	bundle.LoadMessageFileFS(localeFS, "active.en.toml")
}

type C = i18n.LocalizeConfig
type M = i18n.Message

func T(lang string, c C) string {
	s, _ := i18n.NewLocalizer(bundle, lang).Localize(&c)
	return s
}
```

and i can use it like 
```
import "my/wrapper/path/i18n"

...

i18n.T(i18n.C{DefaultMessage:&i18n.M{blah blah blah}})

### DIFF
--- a/v2/goi18n/extract_command.go
+++ b/v2/goi18n/extract_command.go
@@ -38,14 +38,25 @@ Flags:
 		Output message files in this format.
 		Supported formats: json, toml, yaml
 		Default: toml
+
+	-packagepath name
+        Import path of package which contains Message and LocalizeConfig types
+		Default: github.com/nicksnyder/go-i18n/v2/i18n
+
+	-messagetype name
+        Message type name for extracting messages
+		Default: github.com/nicksnyder/go-i18n/v2/i18n
+
 `)
 }
 
 type extractCommand struct {
-	paths          []string
-	sourceLanguage languageTag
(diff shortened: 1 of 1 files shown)


nicksnyder:
This change is just to enable you to write something like `i18n.M` instead of `i18n.Message`?

I am currently hesitant to merge a change like this, it seems unnecessary, and I am wondering if there is a better way to accomplish what you want (i.e. having i18n expose a type alias), but any case a change like this would needs tests before I could consider merging it.

mr-tron:
It allows to make proxy-package inside my project. Aliases is only one example of usage.

nicksnyder:
Closing due to lack of activity.

PR #298, opened 2023-05-09
fix misleading comment on MustLoadMessageFile function

ardaozceviz:
This PR fixes the misleading comment  on `MustLoadMessageFile` function.


nicksnyder:
Thank you!

PR #304, opened 2023-10-05
extract: support typed string constants

hectorj-thetreep:


hectorj-thetreep:
Hi! Here is a minor feature I need for our "enums".

The test case should explain it.

nicksnyder:
Thank you for the contribution and including a test case!

hectorj-thetreep:
Thank you for that quick merge!

Have a good day

PR #189, opened 2019-06-16
simpler fallback behavior

nicksnyder:
Fixes https://github.com/nicksnyder/go-i18n/issues/182

```
name                                                                                        old time/op    new time/op    delta
Localizer_Localize/missing_translation_from_default_language_with_other_translation-8         34.4µs ± 1%     0.3µs ± 6%   -99.11%  (p=0.000 n=8+8)
Localizer_Localize/missing_translation_not_default_language_with_other_translation-8          35.5µs ± 3%     0.4µs ± 9%   -99.00%  (p=0.000 n=8+8)
Localizer_Localize/empty_translation_without_fallback-8                                       32.4µs ± 1%     0.4µs ± 4%   -98.83%  (p=0.001 n=7+7)
Localizer_Localize/message_id_not_mismatched-8                                                 305ns ±22%     279ns ± 4%    -8.45%  (p=0.040 n=8+8)
Localizer_Localize/template_data,_plural_count_float,_default_message-8                       13.1µs ±15%    12.1µs ± 3%    -7.90%  (p=0.002 n=8+8)
Localizer_Localize/template_data,_custom_delims,_default_message-8                            9.21µs ± 2%    8.51µs ± 1%    -7.65%  (p=0.000 n=7+8)
Localizer_Localize/accept_not_default_language,_other_message_in_bundle,_default_message-8     227ns ± 4%     214ns ± 5%    -5.94%  (p=0.001 n=8+8)
Localizer_Localize/template_data,_plural_count_other,_default_message-8                       12.2µs ± 5%    11.5µs ± 1%    -5.84%  (p=0.000 n=7+8)
Localizer_Localize/accept_not_default_language,_message_in_bundle-8                            295ns ± 3%     278ns ± 4%    -5.51%  (p=0.000 n=8+7)
Localizer_Localize/template_data,_plural_count_one,_default_message-8                         12.3µs ± 4%    11.7µs ± 6%    -4.17%  (p=0.029 n=7+8)
Localizer_Localize/template_data,_custom_delims,_bundle_message-8                              821ns ± 3%     795ns ± 5%    -3.13%  (p=0.017 n=7+7)
Localizer_Localize/fallback_to_non-default_less_specific_language-8                            766ns ± 2%     753ns ± 3%    -1.71%  (p=0.036 n=8+8)
Localizer_Localize/missing_translation_from_default_language-8                                 283ns ±11%     285ns ± 2%      ~     (p=0.692 n=8+7)
[...]

PR #305, opened 2023-10-21
Fix Localize regression with empty default message

KuSh:
Behavior has changed with #189. Before that, an empty default message returned nil and an error, now it returns "" without error.

This has been detected by mattermost test suit in https://github.com/mattermost/mattermost/pull/21327 which tries to switch all code base to go-i18n v2, see [related discussion](https://github.com/mattermost/mattermost/pull/21327/files#r1365634307)

### DIFF
--- a/v2/i18n/localizer_test.go
+++ b/v2/i18n/localizer_test.go
@@ -577,6 +577,15 @@ func localizerTests() []localizerTest {
 			},
 			expectedErr: &MessageNotFoundErr{tag: language.AmericanEnglish, messageID: "Hello"},
 		},
+		{
+			name:            "empty default message",
+			defaultLanguage: language.English,
+			acceptLangs:     []string{"en"},
+			conf: &LocalizeConfig{
+				DefaultMessage: &Message{},
+			},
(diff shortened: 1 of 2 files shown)


v2/i18n/localizer_test.go
```diff
@@ -577,6 +577,15 @@ func localizerTests() []localizerTest {
...
 			expectedErr: &MessageNotFoundErr{tag: language.AmericanEnglish, messageID: "Hello"},
 		},
+		{
+			name:            "empty default message",
+			defaultLanguage: language.English,
+			acceptLangs:     []string{"en"},
+			conf: &LocalizeConfig{
+				DefaultMessage: &Message{},
+			},
+			expectedErr: &MessageNotFoundErr{tag: language.English, messageID: ""},
```
nicksnyder, line 587: Can you update the test to use a non-empty message ID?

KuSh: Sure but that won't be exactly what mattermost is testing for. I don't know if that can matter in case of a code revamp

nicksnyder: Could add two test cases if you want

nicksnyder:
Thanks for the fix!

PR #316, opened 2024-01-27
Add option to skip template execution

nohajc:
Hi,

I use your library in production and it works great. Except now I've come to a point where I need it to play nicely with a different string parameter syntax.

Just for context, I'm dealing with a shared repository of localized strings which has multiple consumers using different programming languages and libraries. This already involves conversion from xliff to a format suitable for go-i18n and some of the strings contain placeholders with a C#-like (but simplified) format string syntax (e.g. `"arg1: {0}, arg2: {1}, {{escaped}}"`).

The problem is with escaped curly braces which correspond to parameters in go templates. When I try to localize a string like this, I get en error because it's not even a valid go template.

Therefore it would be very useful if I could just disable template execution and process any string parameters with custom syntax myself.

### DIFF
--- a/v2/i18n/localizer.go
+++ b/v2/i18n/localizer.go
@@ -68,6 +68,10 @@ type LocalizeConfig struct {
 
 	// Funcs is used to extend the Go template engine's built in functions
 	Funcs template.FuncMap
+
+	// NoTemplateExec is used to disable template execution.
+	// The localized string will be returned as is.
+	NoTemplateExec bool
 }
 
 type invalidPluralCountErr struct {
@@ -152,7 +156,7 @@ func (l *Localizer) LocalizeWithTag(lc *LocalizeConfig) (string, language.Tag, e
 	}
 
 	pluralForm := l.pluralForm(tag, operands)
-	msg, err2 := template.Execute(pluralForm, templateData, lc.Funcs)
+	msg, err2 := getOrExecute(template, pluralForm, templateData, lc)
 	if err2 != nil {
 		if err == nil {
 			err = err2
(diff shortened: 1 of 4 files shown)


nohajc:
Btw, I did figure out it's possible to configure custom parameter delimiters, but that still requires using go-style named parameters. This just doesn't work for numeric indices unless I escape any occurrences of double curly braces which basically means turning every `{{` into `{{ "{{" }}` (and the same with `}}`). I'd say it's better to make the library more configurable rather than using workarounds like this.

nicksnyder:
Thanks for the clear problem statement and proposed solution. Something I want to consider is making the templating engine itself customizable. Then it would be possible to provide a no-op templating engineer. I am not sure if this can be done in a backward compatible way, but if it can that might be preferable.

nohajc:
@nicksnyder Would you be willing to merge this before going forward with the more general solution? It would really help me right now... Otherwise I'd probably end up forking the library for the time being.

I mean, the change in public API is rather minimal, it's just one more config option. Of course the no-op templating engine would be a different option but I suppose by then you could just deprecate this one.

nicksnyder:
There are a few issues that would be solved by making the templating system more flexible, so if we are going to add to the public API, I would prefer to do it in a way that will resolve multiple issues at once. I think I have a solution coded up in https://github.com/nicksnyder/go-i18n/pull/317 but I still need to add some tests to verify. 

With that change, instead of `NoTemplateExec = true` you would use `TemplateEngine: NoTemplateEngine{}`.

Would that work for you?

nohajc:
You're fast. :) Definitely, that would work.

nicksnyder:
With 2.4.0 there is now a way to disable template parsing. Check out the [release notes](https://github.com/nicksnyder/go-i18n/releases/tag/v2.4.0) for more info.

PR #176, opened 2019-05-19
v2 only

nicksnyder:
fixes https://github.com/nicksnyder/go-i18n/issues/172

PR #318, opened 2024-01-31
Remove v2 directory

nicksnyder:
The v2 directory is an artifact of the fact that v1 and v2 co-existed in this repository for a while. v1 code was removed in https://github.com/nicksnyder/go-i18n/pull/176. This moves all code that was in the v2 directory to the top level.


jerome-laforge:
It is not conform with semver and it is a big breaking change.

nicksnyder:
Let's discuss in https://github.com/nicksnyder/go-i18n/issues/319

Issue #209, opened 2020-01-10
translation file with key "id" leads LoadMessageFile to fail silently


davleb:
Context : every toml file with a key to translate like
"id" = "Id"

LoadMessageFile returns no error but every use of Localize function  returns an empty string (like it would be if there was an error in LoadMessageFile)
The problem seems to be the identifier "id" itself. If i change it to "ed" or "ad", the problem goes away.
Is it a magic string used internally ? I use the latest V2 version of the lib.

MJacred:
Hi,
this is due to the implementation of `func isMessage(v interface{}) bool`
https://github.com/nicksnyder/go-i18n/blob/26334ab1db070c2801ba26550d7d6238f6c9da10/v2/i18n/message.go#L193-L204
your `"id" = "Id"` is identified as a reserved key and `isMessage` returns `true` and your value is not added as the translation text -> it is blank

The reserved key `id` is needed for json and yaml.  
it would be possible to use it in toml, but `isMessage` is a generic function which acts the same independent of your file format. I don't recommend a change here, as it would have an unnecessary impact on readability and performance.

There's a pull request #208 pending, which would give you an error message in the terminal when you try doing what you did while running the `merge` command:  
`missing translation for message id "id"`  
This should suffice
@nicksnyder: #208 should be fine as-is, no?

davleb:
Hi MJacred,
Thanks for your explanation. I was not aware of the side-effect when using these keys:
"id", "description", "hash", "leftdelim", "rightdelim", "zero", "one", "two", "few", "many", "other"
Not a big deal but if any of these words is used in the toml/yaml file as a key, the bundle is not functional and Localize will return "" for every key, including valid keys. I think an error should be returned at the LoadMessageFile, as it is the case when a key is defined more than one time in the toml/yaml file:
`Near line 14 (last key parsed 'Validate changes'): Key 'Validate changes' has already been defined.`

jeremy-serenne:
I just encountered this issue, it was hard to find this topic as I got the problem with the key "other".

Even if it's a 2 years old issue, do you know when there will be an error returned by `LoadMessageFile` and `LoadMessageFileFS`. I really think it would be very useful for debugging :pray: (@MJacred? I don't know who to ping :sweat_smile:)

andig:
This broke https://github.com/nicksnyder/go-i18n/issues/360. It seems arbitrary that certain TOML keys will now error (or not get processed at all silently).

PR #320, opened 2024-02-04
Error if reserved message keys are mixed with unreserved keys

nicksnyder:
Makes it an error to provide data that mixes reserved message keys with unreserved ones. Previously the unreserved keys would be ignored, but this caused surprising behavior where if someone added a message that had a key with a reserved word, it would silently break the whole translation file. This change turns that situation into an error.

Resolves https://github.com/nicksnyder/go-i18n/issues/209


andriiberezhynskyi:
Hello @nicksnyder 
I did this change 3 years ago:

```diff
-      "descr": "{{if .DeliveryTime}}Livraison estimée : {{.DeliveryTime}} jours{{end}}",
+      "descr": {
+        "other": "{{if .DeliveryTime}}Livraison estimée : {{.DeliveryTime}} jours{{end}}",
+        "leftDelim": "<<",
+        "rightDelim": ">>"
+      },
```
I don't even remember why I did it and don't understand anymore how it all works. Looks like some workaround to support `{{}}` in translation. The problem is that this pr introduces a breaking change and all of a sudden I have this error:
```
reserved keys [other] mixed with unreserved keys [leftDelim rightDelim]
```
How do I fix this problem?

PR #327, opened 2024-05-27
Adds ability to extract strings from T() functions to `goi18n extract`

allmightyspiff:
### Purpose
`goi18n extract` will now extract strings wrapped in a `T` function, in addition to the other cases it extracts strings from.
I also added an example CLI application to go with the web application that shows how to embed i18n files, and use T wrapper functions for translations.

### Reason
This feature will help people like me migrating from v1 to v2 of this library. V1 encouraged the use of the `T` wrapper, so having a way to easily extract those strings would make my life a lot easier, and I assume others who make the migration as well.

Thanks for your time reviewing this issue, and I'll address any issues you may have with the code.
Also thanks for making this library in the first place.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,4 +1,4 @@
-# Example
+# Web Example
 
 This directory contains an example project that uses go-i18n.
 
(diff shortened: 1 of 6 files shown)


nicksnyder:
I appreciate the effort, but this contribution is not something I want to be on the hook of maintaining.

PR #335, opened 2024-10-08
fix: option in TextParser is not used

### LINKED ISSUE #334: The Option in TextParser is not used
If we define a custom TemplateParser with some Option, the option has no effect.

Example:

```go
bundle := NewBundle(language.English)
bundle.AddMessages(language.English, {ID: "Foo", Other: "Foo {{.bar}}!"}...)
conf := &LocalizeConfig{
  MessageID:      "Foo",
  TemplateData:   map[string]string{}, // empty data
  TemplateParser: &template.TextParser{Option: "missingkey=error"}, // I want an error if `bar` is not in data
}
localizer := NewLocalizer(bundle, []string{"en"})
s, err := localizer.Localize(conf)) // I expect an error here

```

This is because the `.Option()` is not used in TextParser.

cazeaux:
Resolves #334

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,12 +4,6 @@ go 1.18
 
 require (
 	github.com/BurntSushi/toml v1.4.0
-	github.com/stretchr/testify v1.9.0
 	golang.org/x/text v0.17.0
 	gopkg.in/yaml.v3 v3.0.1
-)
-
-require (
-	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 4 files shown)


nicksnyder:
Ugh, thanks for finding this. `Option` in `TextParser` really should have been a `[]string`, but changing now would be backwards incompatible. There is only one supported option today, so this is moot for now.

I want to avoid pulling in an extra library just for one test so am going to push some changes.

PR #336, opened 2024-10-12
Create README.uk-UA.md

dejurin:


nicksnyder:
I added a section in the main README about translations of this document. Can you add and translate this section?

https://github.com/nicksnyder/go-i18n?tab=readme-ov-file#translations-of-this-document

PR #418, opened 2026-09-18
i18n: keep empty message strings on import

### LINKED ISSUE #329: Empty string as message content is ignored
I use JSON files for translations, like

```json
{
  "application: {
    "formatter": {
      "name": "My application formatter",
      "desc": "Some explanatory description for the formatter"
    },
    "v1": {
      "label": "Label for the version v1",
      "desc": "Description for this version",
      "fields": {
        "FirstVariableToSet": {
          "label": "MyVariable",
          "desc": ""
        }
      }
    }
  }
}
```

I think it is absolutely valid to have an empty translation for a key, like `application.v1.fields.FirstVariableToSet.desc`.

Unfortunately the key (and its value) is completely scrapped during message file import. 

Culprit: https://github.com/nicksnyder/go-i18n/blob/b42d9bd0c568eebf5ce123ec94b9b13438f46855/i18n/message_template.go#L37

Would it be possible to support empty message values in the future? Or maybe with a configuration parameter, whether they should be ignored or not.

team-humaki:
Fixes #329

Empty translation values (e.g. `"desc": ""` in nested JSON) were parsed into messages, then dropped when adding them to a bundle because `NewMessageTemplate` skips blank plural forms. `AddMessages` now keeps forms that were present in the source, including an explicit empty string, so `Localize` finds the key and returns `""`.

`NewMessageTemplate` is unchanged, so goi18n extract/merge still treat blank forms as omitted.


nicksnyder:
The original issue doesn't contain a justification for why it is useful to keep translation with empty message keys.

ccoVeille:
This account spams several repositories, including the one I maintain.

Look at the contributions of the user.

I would simply ignore this PR if I was you @nicksnyder

PR #419, opened 2026-09-18
i18n: accept language tag for directory-based message files

### LINKED ISSUE #347: Allow MustLoadMessageFile to accept an optional langTag
`MustLoadMessageFile()` takes a path that follows the file naming convention of `xxx.en.json` and automatically parses the language tag. This incorrectly retrieves the language tag if the path does not follow the convention. In my case, the locales path is `locales/en/translation.json`. Passing this path to `MustLoadMessageFile()` will result in the langTag being `translation` instead of `en`. It'd be super helpful if either the function could also parse the aforementioned path or accept a langTag to override the automatic parsing.

team-humaki:
Fixes #347.

`MustLoadMessageFile` inferred the language from the filename, so `locales/en/translation.json` was treated as tag `translation`.

- Infer a well-formed parent directory when the filename is not a language tag.
- Add `LoadMessageFileWithTag` / `MustLoadMessageFileWithTag` / `ParseMessageFileBytesWithTag` / `LoadMessageFileFSWithTag` to override the path.


i18n/bundle.go
```diff
@@ -67,16 +67,50 @@ func (b *Bundle) MustLoadMessageFile(path string) {
...
+	}
+}
+
 // ParseMessageFileBytes parses the bytes in buf to add translations to the bundle.
 //
 // The format of the file is everything after the last ".".
 //
 // The language tag of the file is everything after the second to last "." or after the last path separator, but before the format.
+// If that name is not a well-formed language tag, a well-formed parent directory is used
+// (for example locales/en/translation.json).
```
nicksnyder: This comment appears to be incorrect
```suggestion
```

i18n/parse.go
```diff
@@ -157,6 +157,23 @@ func addChildMessages(id string, data interface{}, messages []*Message) ([]*Mess
...
+	}
+	parent := filepath.Base(filepath.Dir(path))
+	if parent != "" && parent != "." && parent != string(os.PathSeparator) {
+		if tag, err := language.Parse(parent); err == nil {
+			return tag
+		}
+	}
+	return language.Make(lang)
+}
+
```
nicksnyder: I am not comfortable making this change. We should not further increase the number of magic cases, the correct API is to explicitly provide the tag.

i18n/bundle.go
team-humaki: Dropped that parent-directory sentence. ParseMessageFileBytes still infers the tag from the filename only, as before.

i18n/parse.go
team-humaki: Removed tagFromPath. Path inference is filename-only again; directory-based files go through LoadMessageFileWithTag / ParseMessageFileBytesWithTag.

